PR #46, opened 2018-05-31
Smooth slide-up animation of Event Cards now occurs

dreadpool2:
32957146_1735272249841408_463897648114434048_n

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
@@ -18,6 +18,9 @@ class EventsFragment : Fragment() {
     private val eventsRecyclerAdapter: EventsRecyclerAdapter = EventsRecyclerAdapter()
     private val eventsViewModel by viewModel<EventsViewModel>()
     private lateinit var rootView: View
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -18,6 +18,9 @@ class EventsFragment : Fragment() {
     private val eventsRecyclerAdapter: EventsRecyclerAdapter = EventsRecyclerAdapter()
     private val eventsViewModel by viewModel<EventsViewModel>()
     private lateinit var rootView: View
+    private val linearLayoutManager: LinearLayoutManager by lazy {
+        LinearLayoutManager(activity)
+    }
```
iamareebjamal: App will crash on back press

dreadpool2: Updated the PR

iamareebjamal: This should be lateinit var

dreadpool2:
Updated PR

PR #52, opened 2018-06-01
feat: Add DateTime feature and test sharing content

iamareebjamal:


app/src/main/java/org/fossasia/openevent/general/event/EventUtils.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        message.append(resource.getString(R.string.event_name)).append(event.name).append("\n\n")
+        if (!description.isEmpty()) message.append(resource.getString(R.string.event_description))
+                .append(event.description.nullToEmpty()).append("\n\n")
+        message.append(resource.getString(R.string.starts_on))
+                .append(startsAt.format(dateFormat)).append(" ")
+                .append(startsAt.format(timeFormat)).append("\n")
+        message.append(resource.getString(R.string.ends_on))
+                .append(endsAt.format(dateFormat)).append(" ")
+                .append(endsAt.format(timeFormat))
```
dreadpool2: @iamareebjamal So we are assuming that there will be always a start date and an end date?

iamareebjamal: Because the API says so

How can there be an event without start and end date?

Check API documentation

dreadpool2: @iamareebjamal Yeah of course there can't be :+1:

PR #74, opened 2018-06-04
feat: Add events into database

### LINKED ISSUE #62: Implement local caching of events in room
working on it 👍

iamareebjamal:
Fixes #62


iamareebjamal:
@nikit19 @dreadpool2 @simarsingh24 Please see the approach, review and make suggestions if and how it can be improved

iamareebjamal:
@nikit19 Is it clear now why you weren't able to inject 

Also, please tell if you understand the logic of loading the events from DB and network in service

nikit19:
Yes I think I understand the logic.
For the first time it will load the events from the network and then it will check if the database is empty or not. Right ?

simarsingh24:
Thanks! Will follow the same approach for #62

PR #80, opened 2018-06-05
feat : Add users into database

### LINKED ISSUE #63: implement local caching of users in room
working on it 👍 
to be done after #59

simarsingh24:
fixes #63

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
@@ -2,6 +2,9 @@ package org.fossasia.openevent.general.auth
 
 import io.reactivex.Flowable
 import io.reactivex.Single
+import io.reactivex.android.schedulers.AndroidSchedulers
+import io.reactivex.schedulers.Schedulers
+import timber.log.Timber
 
 
 class AuthService(private val authApi: AuthApi,
@@ -33,7 +36,16 @@ class AuthService(private val authApi: AuthApi,
     fun logout() {
         val userFlowable = userDao.getUser(authHolder.getId())
         userFlowable.map {
-            userDao.deleteUser(it)
+            Single.fromCallable {
+                userDao.deleteUser(it)
+            }.subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe({
+                        Timber.d("User logged out!")
+                    }
+                    ) {
+                        Timber.e(it, "Error logging out user!")
+                    }
         }
         authHolder.token = null
     }
@@ -41,19 +53,18 @@ class AuthService(private val authApi: AuthApi,
 
     fun getProfile(id: Long = authHolder.getId()): Flowable<User> {
         val userFlowable = userDao.getUser(id)
-        return userFlowable.switchMap {
-            if (it != null)
-                userFlowable
-            else
-                authApi.getProfile(id)
-                        .map {
-                            userDao.insertUser(it)
-                        }
-                        .toFlowable()
-                        .flatMap {
-                            userFlowable
-                        }
+        userFlowable.switchIfEmpty{
+            authApi.getProfile(id)
+                    .map {
+                        userDao.insertUser(it)
+                    }
+                    .toFlowable()
+                    .flatMap {
+                        userFlowable
+                    }
         }
+        return userFlowable
+        
     }
 
 }
--- a/app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
@@ -1,10 +1,7 @@
 package org.fossasia.openevent.general.auth
 
-import io.reactivex.Flowable
+import io.reactivex.Completable
 import io.reactivex.Single
-import io.reactivex.android.schedulers.AndroidSchedulers
-import io.reactivex.schedulers.Schedulers
-import timber.log.Timber
 
 
 class AuthService(private val authApi: AuthApi,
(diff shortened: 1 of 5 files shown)


simarsingh24:
please check it thoroughly it was not working on my phone, not sure why

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -33,21 +31,29 @@ class AuthService(private val authApi: AuthApi,
     fun isLoggedIn() = authHolder.isLoggedIn()
 
     fun logout() {
-        userMap.remove(authHolder.getId())
-
+        val userFlowable = userDao.getUser(authHolder.getId())
+        userFlowable.map {
+            userDao.deleteUser(it)
+        }
```
iamareebjamal: This doesn't get subscribed anywhere, how will it execute

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -33,21 +31,29 @@ class AuthService(private val authApi: AuthApi,
...
 
-        return authApi.getProfile(id)
-                .map {
-                    userMap[id] = it
-                    it
-                }
+    fun getProfile(id: Long = authHolder.getId()): Flowable<User> {
+        val userFlowable = userDao.getUser(id)
+        return userFlowable.switchMap {
+            if (it != null)
```
iamareebjamal: This will never be null

simarsingh24:
still not working

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -33,21 +34,37 @@ class AuthService(private val authApi: AuthApi,
...
+                userDao.deleteUser(it)
+            }.subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe({
+                        Timber.d("User logged out!")
+                    }
+                    ) {
+                        Timber.e(it, "Error logging out user!")
+                    }
+        }
```
iamareebjamal: Logic is not correct, also, return the completable from here and subscribe on the caller site

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -33,21 +34,37 @@ class AuthService(private val authApi: AuthApi,
...
-            return Single.just(user)
 
-        return authApi.getProfile(id)
-                .map {
-                    userMap[id] = it
-                    it
-                }
+    fun getProfile(id: Long = authHolder.getId()): Flowable<User> {
+        val userFlowable = userDao.getUser(id)
+        userFlowable.switchIfEmpty{
```
iamareebjamal: This will never be empty

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
simarsingh24: okay

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
simarsingh24: what should be the condition to fetch user from network then? (when user not found from local db)

iamareebjamal: When user is not present, an error is thrown, verify that this happens. If so, use onErrorResumeNext

simarsingh24: tried inside(getProfile())
```userFlowable.doOnError{
            Timber.e("Error fetching user")
        }
```
but no error was shown in log trace

iamareebjamal: What happened when you subscribed it. RxJava guarantees that you'll either get data or error. If you didn't get error, you must have gotten data, and if there is no user with that ID, you couldn't have

simarsingh24:
screen shot 2018-06-07 at 6 59 14 am

simarsingh24:
changed the query from `flowable<user>` to `maybe<user>` and it was working maybe `completes` if it couldnt find any user in database

simarsingh24:
or we can user single<user> also in that case onError is called if no user is found

simarsingh24:
@iamareebjamal what are your views

iamareebjamal:
Use Single

simarsingh24:
okay

simarsingh24:
@iamareebjamal still not working please look into it, tried many ways to do it :(

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -33,39 +30,31 @@ class AuthService(private val authApi: AuthApi,
...
-                    .observeOn(AndroidSchedulers.mainThread())
-                    .subscribe({
-                        Timber.d("User logged out!")
-                    }
-                    ) {
-                        Timber.e(it, "Error logging out user!")
-                    }
-        }
-        authHolder.token = null
+            }
```
iamareebjamal: Won't work as there is nobody subscribing it

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -33,39 +30,31 @@ class AuthService(private val authApi: AuthApi,
...
-                        userFlowable
-                    }
+    fun getProfile(id: Long = authHolder.getId()): Single<User> {
+        val userSingle = userDao.getUser(id)
+        userSingle.onErrorResumeNext {
+            var profileSingle = authApi.getProfile(id)
+            profileSingle.map {
+                userDao.insertUser(it)
+            }
+            return@onErrorResumeNext profileSingle
```
iamareebjamal: Return directly without storing in intermediate variable

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
simarsingh24: should I then keep it inside above map statement?

iamareebjamal: You can

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
simarsingh24: one map statement also has to be executed that's why intermediated variable is added

simarsingh24:
onErrorResumeNext block is not getting executed

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
iamareebjamal: Can be done with implicit returning as well

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -35,23 +35,21 @@ class AuthService(private val authApi: AuthApi,
...
-            return@onErrorResumeNext profileSingle
-        }
-        return userSingle
+        return userDao.getUser(id)
+                .onErrorResumeNext {
+                    var profileSingle = authApi.getProfile(id)
+                    profileSingle.map {
+                        userDao.insertUser(it)
+                    }
+                    return@onErrorResumeNext profileSingle
```
iamareebjamal: Use direct return

simarsingh24:
screen shot 2018-06-09 at 5 43 05 pm
@iamareebjamal  not allowed

iamareebjamal:
Once you remove the redundant variable holder, it will be

iamareebjamal:
```
.onErrorResumeNext {
    authApi.getProfile(id)
        .map {
            userDao.insertUsers(it)
            it
        }
    }
}
```

simarsingh24:
screen shot 2018-06-09 at 5 47 03 pm

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -32,23 +30,29 @@ class AuthService(private val authApi: AuthApi,
...
+                    authApi.getProfile(id)
+                            .map {
+                                userDao.insertUser(it)
+                                it
+                            }
                 }
+
     }
 
+
```
nikit19: Please remove extra blank lines

iamareebjamal:
@simarsingh24 Please check the changed logic I pushed

simarsingh24:
@iamareebjamal LGTM 👍

nikit19:
@iamareebjamal This PR got merged to the master branch. Could this be a problem ? There is also one more PR which got merged in the master branch

iamareebjamal:
Yes, we have to synchronize it now. Doing so. From now, target dev branch

PR #93, opened 2018-06-07
fix: event url added in share

### LINKED ISSUE #92: Share FAB should have the url of the event
It should not contain other details just the URL of the event hosted on eventyay
Working on it

nikit19:
fixes #92 
screenshot_20180607-120630

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventUtils.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventUtils.kt
@@ -16,10 +16,12 @@ object EventUtils {
 
     private val timeFormat: DateTimeFormatter = DateTimeFormatter.ofPattern("hh:mm a")
     private val dateFormat: DateTimeFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy")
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventUtils.kt
```diff
@@ -16,10 +16,12 @@ object EventUtils {
 
     private val timeFormat: DateTimeFormatter = DateTimeFormatter.ofPattern("hh:mm a")
     private val dateFormat: DateTimeFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy")
+    private val frontendUrl: String = "https://open-event-frontend-dev.herokuapp.com/e/"
```
iamareebjamal: This is dev environment. Not necessary the event will be there

nikit19: Actually in today's meeting we were asked to change it to this one. I also tested it manually for some events and the links were always working. What do you suggest should be done ?

iamareebjamal: Then it's fine

PR #99, opened 2018-06-07
chore: reformat code and optimize imports

krupalshah:


app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragmentViewModel.kt
```diff
@@ -26,8 +26,8 @@ class ProfileFragmentViewModel(private val authService: AuthService) : ViewModel
                 .doOnSubscribe({
                     progress.value = true
                 }).doFinally {
-                    progress.value = false
-                }.subscribe({ user ->
+            progress.value = false
+        }.subscribe({ user ->
```
iamareebjamal: This is wrongly indented

krupalshah: @iamareebjamal It seems to be right. I just run reformat code on whole package.

PR #112, opened 2018-06-10
feat: Make search queries case insensitive

### LINKED ISSUE #110: Search results should not depend on case of the query
It should be case insensitive. Currently on searching "Test" and "test" will give different results this should not be the case.
working on it

nikit19:
fixes #110


iamareebjamal:
Please follow commit guidelines

nikit19:
commit message updated

nikit19:
@iamareebjamal could you merge this if no further changes are required

PR #116, opened 2018-06-11
fix: do not open events fragment on screen rotation

### LINKED ISSUE #115: Events fragment get opened on screen rotation
Every time the screen rotates in Main Activity (search/profile fragment),event fragment opens up
working on it

nikit19:
fixes #115


app/src/main/AndroidManifest.xml
```diff
@@ -12,7 +12,9 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
-        <activity android:name=".MainActivity">
+        <activity
+            android:name=".MainActivity"
+            android:configChanges="orientation|screenSize">
```
iamareebjamal: Please don't lock the orientation. You don't kill the patient when he has a disease

nikit19: Actually this does not lock the orientation. It is just that that activity does not get re created on configuration changes.

simarsingh24: location is not locked!

iamareebjamal: Which means that the orientation of layout is locked. If you implement tablet layout in future, it won't be applied on rotation. This is not correct behaviour since we are already using ViewModel and other lifecycle aware components

iamareebjamal: Even the screen size is locked which will create problems with Android Multi Tasking view and PiP

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -45,7 +47,10 @@ class MainActivity : AppCompatActivity() {
 
         supportActionBar?.title = "Events"
 
-        loadFragment(EventsFragment())
+        if(mainActivityViewModel.loadEventsAgain.value==null) {
+            loadFragment(EventsFragment())
+            mainActivityViewModel.loadEventsAgain.value=false
+        }
```
iamareebjamal: Uhmm, just check if savedInstanceState is null

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -48,10 +45,9 @@ class MainActivity : AppCompatActivity() {
 
         supportActionBar?.title = "Events"
 
-        if (mainActivityViewModel.loadEventsAgain.value == null) {
-            loadFragment(EventsFragment())
-            mainActivityViewModel.loadEventsAgain.value = false
-        }
+       if(savedInstanceState==null)
```
iamareebjamal: Space around symbols

app/src/main/AndroidManifest.xml
```diff
@@ -23,7 +23,6 @@
             android:name="org.fossasia.openevent.general.AuthActivity"
             android:windowSoftInputMode="adjustPan">
         </activity>
-        <activity android:name=".auth.LoginActivity"></activity>
```
iamareebjamal: Unneeded

nikit19:
PR updated Please review

PR #141, opened 2018-06-16
feat: get events by location in home screen

### LINKED ISSUE #78: Get events by location in the Home Screen
Something like this
screenshot_20180605-013433

nikit19:
fixes #78 
ezgif com-video-to-gif 7

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
@@ -30,6 +30,13 @@ class EventService(private val eventApi: EventApi, private val eventDao: EventDa
                 }
     }
 
+    fun getEventsByLocation(locationName: String): Single<List<Event>> {
+        return eventApi.searchEvents("name", locationName).map {
+            eventDao.insertEvents(it)
+            it
+        }
+    }
+
     fun getEvent(id: Long): Flowable<Event> {
         return eventDao.getEvent(id)
     }
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
@@ -31,10 +31,7 @@ class EventService(private val eventApi: EventApi, private val eventDao: EventDa
     }
 
     fun getEventsByLocation(locationName: String): Single<List<Event>> {
-        return eventApi.searchEvents("name", locationName).map {
-            eventDao.insertEvents(it)
-            it
-        }
+        return eventApi.searchEvents("name", locationName)
     }
 
     fun getEvent(id: Long): Flowable<Event> {
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
@@ -31,7 +31,10 @@ class EventService(private val eventApi: EventApi, private val eventDao: EventDa
     }
 
     fun getEventsByLocation(locationName: String): Single<List<Event>> {
-        return eventApi.searchEvents("name", locationName)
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -69,7 +75,23 @@ class EventsFragment : Fragment() {
             it?.let { showProgressBar(it) }
         })
 
-        eventsViewModel.loadEvents()
+        /*if (eventsViewModel.showAllEvents) {
+            eventsViewModel.loadEvents()
+            eventsViewModel.showAllEvents = false
+        }*/
```
nikit19: @iamareebjamal All events are loaded if I uncomment this when the app starts(as expected). But when I put the location of the event and search then if the event is already present in the list of events, all the events in the table are displayed  otherwise if I am searching the event for the first time correct events are displayed.

So what I am doing right now is just searching the events according to location when the user enters the location similar to how it is happening in the Search fragment. I am a little confused about  what should be the correct approach here

nikit19: Yes always

iamareebjamal: So that means they may also always launch online query and never show events from database?

nikit19: Yes exactly ! I will update the PR with the required changes

nikit19:
@iamareebjamal I have updated the PR, I am fetching the events directly from the network.
Please Review

app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
```diff
@@ -31,10 +31,7 @@ class EventService(private val eventApi: EventApi, private val eventDao: EventDa
     }
 
     fun getEventsByLocation(locationName: String): Single<List<Event>> {
-        return eventApi.searchEvents("name", locationName).map {
-            eventDao.insertEvents(it)
```
iamareebjamal: You have to still save them in DB

nikit19:
@iamareebjamal PR updated. Please Review

app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -69,7 +75,18 @@ class EventsFragment : Fragment() {
...
-        eventsViewModel.loadEvents()
+        rootView.locationEdittext.setOnEditorActionListener(TextView.OnEditorActionListener { v, actionId, event ->
+            if (actionId == EditorInfo.IME_ACTION_SEARCH && !TextUtils.isEmpty(rootView.locationEdittext.text)) {
+                val imm = context?.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+                imm.hideSoftInputFromWindow(rootView.locationEdittext.windowToken, 0)
+
+                eventsViewModel.locationName = rootView.locationEdittext.text.toString()
+                eventsViewModel.loadEventsWithAnimation = false
+                eventsViewModel.loadLocationEvents()
+                return@OnEditorActionListener true
```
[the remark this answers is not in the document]
simarsingh24: replace it with normal return statment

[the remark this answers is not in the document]
nikit19: doesn't work. I get an error as soon as I remove `@OnEditorActionListener`
`The boolean literal does not conform to the expected type View?`

nikit19:
@iamareebjamal could you please review this PR whenever you get time
I would like to work on #145 and other issues after this PR gets merged.

PR #143, opened 2018-06-16
fix : corrected query in socialLinkDao

### LINKED ISSUE #142: correct query in socialLinkDAO
working on it 👍

simarsingh24:
fixes #142


app/src/main/java/org/fossasia/openevent/general/social/SocialLinksDao.kt
```diff
@@ -15,5 +16,5 @@ interface SocialLinksDao {
     fun deleteAll()
 
     @Query("SELECT * from SocialLink WHERE event = :eventId")
-    fun getAllSocialLinks(eventId: Long): Flowable<List<SocialLink>>
+    fun getAllSocialLinks(eventId: EventId): Flowable<List<SocialLink>>
```
iamareebjamal: Rename the function appropriately

Also, using long is better

simarsingh24: closing this because there is no error in this query false issue

PR #230, opened 2018-06-30
fix: set default value of qty to zero

### LINKED ISSUE #229: default value of tickets qty spinner should be zero
working on it

nikit19:
fixes #229 
screenshot_20180630-200154


app/src/main/java/org/fossasia/openevent/general/ticket/TicketViewHolder.kt
```diff
@@ -14,6 +14,7 @@ class TicketViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
 
         if (ticket.minOrder > 0 && ticket.maxOrder > 0) {
             val spinnerList = ArrayList<String>()
+            spinnerList.add(Integer.toString(0))
```
iamareebjamal: "0"

nikit19: PR Updated

PR #242, opened 2018-07-02
fix: Favorite FAB drawable bug

### LINKED ISSUE #165: favourite fab drawable not changing on adding to favourite
working on it 👍

simarsingh24:
fixes #165

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventViewHolder.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventViewHolder.kt
@@ -40,7 +40,7 @@ class EventViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
         }
 
         itemView.favoriteFab.setOnClickListener {
-            favoriteListener?.onClick(event.id, event.favorite)
+            favoriteListener?.onClick(event, event.favorite)
         }
     }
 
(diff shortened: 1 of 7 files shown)


simarsingh24:
After change gif
videotogif_2018 07 03_22 54 45
@nikit19 @dreadpool2 please review

iamareebjamal:
This is a workaround. Why is Room not triggering listener once the table content is changed

simarsingh24:
@iamareebjamal We are displaying events from local databse and observer is set on that only when we do setFavorite it updates the event store locally which means automatically recycler views should be updated (ideally), this is the second best solution to this bug I could think of, not sure why observer  is not triggered after update also if it gets called 
```
eventsViewModel.events.observe(this, Observer {
            it?.let {
                eventsRecyclerAdapter.addAll(it)
                eventsRecyclerAdapter.notifyDataSetChanged()
            }
            Timber.d("Fetched events of size %s", eventsRecyclerAdapter.itemCount)
        })
```
then according to this even if a single event is updated all of the events will be inserted once again which should not happen

iamareebjamal:
> even if a single event is updated all of the events will be inserted once again which should not happen
It should happen

There is no point of using Flowable from Room if we're gonna do it this way

There's a reason database is the single source of truth

PR #257, opened 2018-07-04
feat: Exclude same event from Similar Events

### LINKED ISSUE #254: Exclude same event being shown in similar events section
For example, Event Electronics Thunder 2018 includes the same event as a similar event.
Working on it :+1:

dreadpool2:
Fixes #254

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsRecyclerAdapter.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsRecyclerAdapter.kt
@@ -5,7 +5,6 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import org.fossasia.openevent.general.R
-import org.fossasia.openevent.general.event.topic.SimilarEventsFragment
 import java.util.*
 
 class EventsRecyclerAdapter : RecyclerView.Adapter<EventViewHolder>() {
@@ -13,6 +12,8 @@ class EventsRecyclerAdapter : RecyclerView.Adapter<EventViewHolder>() {
     private var clickListener: RecyclerViewClickListener? = null
     private var favoriteFabListener: FavoriteFabListener? = null
     private var eventLayout: String? = null
+    private var idEvent: Long = -1
+    private lateinit var eventView: View
 
     fun setListener(listener: RecyclerViewClickListener) {
         clickListener = listener
@@ -26,6 +27,10 @@ class EventsRecyclerAdapter : RecyclerView.Adapter<EventViewHolder>() {
         eventLayout = type
     }
 
+    fun setEventId(id: Long) {
+        idEvent = id
+    }
+
     fun addAll(eventList: List<Event>) {
         if (events.isNotEmpty())
             this.events.clear()
@@ -33,7 +38,6 @@ class EventsRecyclerAdapter : RecyclerView.Adapter<EventViewHolder>() {
     }
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): EventViewHolder {
-        val eventView: View
         if (eventLayout.equals(SIMILAR_EVENTS)) {
             eventView = LayoutInflater.from(parent.context).inflate(R.layout.item_card_similar_events, parent, false)
         } else {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsRecyclerAdapter.kt
```diff
@@ -5,14 +5,15 @@ import android.view.LayoutInflater
...
 import org.fossasia.openevent.general.R
-import org.fossasia.openevent.general.event.topic.SimilarEventsFragment
 import java.util.*
 
 class EventsRecyclerAdapter : RecyclerView.Adapter<EventViewHolder>() {
     private val events = ArrayList<Event>()
     private var clickListener: RecyclerViewClickListener? = null
     private var favoriteFabListener: FavoriteFabListener? = null
     private var eventLayout: String? = null
+    private var idEvent: Long = -1
```
nikit19: `eventId`

app/src/main/java/org/fossasia/openevent/general/event/EventsRecyclerAdapter.kt
```diff
@@ -44,13 +48,25 @@ class EventsRecyclerAdapter : RecyclerView.Adapter<EventViewHolder>() {
...
+        handleEventVisibility(event, holder)
     }
 
     override fun getItemCount(): Int {
         return events.size
     }
 
+    private fun handleEventVisibility(event: Event, holder: EventViewHolder) {
+        if (eventLayout.equals(SIMILAR_EVENTS)) {
+            if (idEvent == event.id) {
```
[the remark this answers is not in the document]
dreadpool2: PR Updated :+1:

app/src/main/java/org/fossasia/openevent/general/event/topic/SimilarEventsViewModel.kt
```diff
@@ -26,9 +26,8 @@ class SimilarEventsViewModel(private val eventService: EventService) : ViewModel
                 .observeOn(AndroidSchedulers.mainThread())
                 .doOnSubscribe({
                     progress.value = true
-                }).doFinally({
-                    progress.value = false
                 }).subscribe({
+                    progress.value = false
                     similarEvents.value = it
```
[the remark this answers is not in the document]
dreadpool2: PR Updated :+1:

app/src/main/java/org/fossasia/openevent/general/event/topic/SimilarEventsViewModel.kt
```diff
@@ -16,20 +17,24 @@ class SimilarEventsViewModel(private val eventService: EventService) : ViewModel
...
+    var eventId: Long = -1
 
     fun loadSimilarEvents(id: Long) {
         if(id == -1L){
             return
         }
         compositeDisposable.add(eventService.getSimilarEvents(id)
                 .subscribeOn(Schedulers.io())
+                .flatMapIterable{ it }
+                .filter { it.id != eventId }
```
iamareebjamal: Add toList

iamareebjamal: You have to maintain a new list because of it, and new trigger will duplicate items. Logic is wrong

dreadpool2:
PR Updated :+1: . Used switch map and it works fine :+1:

app/src/main/java/org/fossasia/openevent/general/event/topic/SimilarEventsViewModel.kt
```diff
@@ -16,19 +17,25 @@ class SimilarEventsViewModel(private val eventService: EventService) : ViewModel
...
             return
         }
         compositeDisposable.add(eventService.getSimilarEvents(id)
                 .subscribeOn(Schedulers.io())
+                .switchMap {
+                    Observable.fromIterable(it)
+                        .filter{ it.id != eventId }
+                        .toList()
+                        .toFlowable()
+                }
```
iamareebjamal: Not right

dreadpool2: toFlowable shoudn't be used ?

dreadpool2: The logic does satisfy here right as I am using a switch map to filter the event one by one and then converting it to a list and the list is then converted to a flowable. Is the logic still wrong ?

dreadpool2: I researched a bit an found out that the Flowable which is locked to the information returning from the Dao is not suitable for toList. toList requires a finite source. Converting Flowable to Single will be fruitful in this case.

2nd answer states this
https://stackoverflow.com/questions/50318141/rxjava-having-trouble-converting-one-list-to-another-list

iamareebjamal: OK, then it would be more efficient to remove the item in subscribe method than create another stream

dreadpool2: Yup :+1: Let me try that

dreadpool2: Updated PR

PR #264, opened 2018-07-06
fix: correct layout params for favorite event item

### LINKED ISSUE #199: Correct layout parameters of Favourite event item
The problem occurs when the name of event is large. ( The date is not clearly visible )

screenshot_20180625-035758

Working on it :+1:

simarsingh24:
fixes #199


simarsingh24:
Before-
image
After-
screenshot-1530893664248

PR #298, opened 2018-07-10
fix: Add no search results message

### LINKED ISSUE #295: Show no search results message when no search results are found
screenshot_20180710-122729

Working on it :+1:

dreadpool2:
Fixes #295 
videotogif_2018 07 10_21 51 25

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -28,7 +28,7 @@
     <string name="what_time">What time works?</string>
     <string name="pick_date">Pick a date...</string>
     <string name="at">at</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -28,6 +28,7 @@
     <string name="what_time">What time works?</string>
     <string name="pick_date">Pick a date...</string>
     <string name="at">at</string>
+    <string name="no_search_results">No Search Results!\nTry another search</string>
```
nikit19: Can you use the same message that is used in EventBrite. Their message is a little different

dreadpool2: updated message :+1:

PR #316, opened 2018-07-13
feat: Show correct currency in Tickets Fragment

### LINKED ISSUE #311: Show correct currency in Tickets Fragment
Right now it is always $

dreadpool2:
Fixes #311 
screenshot_20180712-213120


app/src/main/java/org/fossasia/openevent/general/utils/CurrencyUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+            try {
+                val currency = Currency.getInstance(locale)
+                currencyLocaleMap.put(currency, locale)
+            } catch (e: NullPointerException) {
+                Timber.d(e, "Failed!")
+            }
+        }
+    }
+
+    fun getCurrencySymbol(currencyCode: String?): String? {
```
nikit19: Same method should be used everywhere to convert into the symbol. 
Is this more accurate than the one used before? If yes use this method in the attendee fragment as well.

dreadpool2: Sure :+1:

app/src/main/java/org/fossasia/openevent/general/utils/CurrencyUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+                currencyLocaleMap.put(currency, locale)
+            } catch (e: IllegalArgumentException) {
+                Timber.d(e, "Failed!")
+            }
+        }
+    }
+
+    fun getCurrencySymbol(currencyCode: String?): String? {
+        val currency = Currency.getInstance(currencyCode)
+        return currency.getSymbol(currencyLocaleMap[currency])
```
nikit19: I think there is no need of creating a util class for this
 `currency.symbol` does the same thing what this entire util class is doing.

dreadpool2: Okay :+1:

dreadpool2: @nikit19 Updated PR

app/src/main/java/org/fossasia/openevent/general/ticket/TicketViewHolder.kt
```diff
@@ -30,9 +30,15 @@ class TicketViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
             itemView.orderRange.adapter = ArrayAdapter(itemView.context, R.layout.select_dialog_singlechoice, spinnerList)
         }
 
+        val price = StringBuilder()
```
nikit19: why not store the currency in a different variable and reference it using $ in `  itemView.price.text`
using the stringBuilder makes it look a little complicated

dreadpool2: Updated

iamareebjamal:
Why pass it around through bundles when it can be calculated?

dreadpool2:
@iamareebjamal Actually the currency used is present in the event object, not present in the ticket object. So I am passing the currency present in the Event object by a bundle to the Tickets fragment

iamareebjamal:
That's not good

dreadpool2:
@iamareebjamal Could you think of any ideal solution to this? I am not able to find any other alternate :(

iamareebjamal:
OK, right now we'll use this. Create an issue to handle this later

dreadpool2:
Okay will do that :+1:

dreadpool2:
@iamareebjamal Created an issue :+1:

nikit19:
@dreadpool2 Looks good to me! could you also convert the currency in ticket details too. Otherwise this is ready to be merged

dreadpool2:
Doing that :+1:

PR #364, opened 2018-07-22
Merge branch 'development'

nikit19:


nikit19:
@iamareebjamal Please review. 
Please tell if I did something wrong
To avoid a merge commit, rebase and merge should be selected right?

nikit19:
@iamareebjamal Please review.
Please tell if I did something wrong
To avoid a merge commit, rebase and merge should be selected right?

iamareebjamal:
Yes, rebase should be selected but rebase sometimes split makes master branch X commits ahead and X commits behind development so I prefer merge commit on master, but if it is desired then rebase should be done

PR #365, opened 2018-07-23
feat: Store user on logging in

### LINKED ISSUE #362: Store user profile on login
Working on it :+1:

dreadpool2:
Fixes #362

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
@@ -47,7 +47,7 @@ class LoginFragment : Fragment() {
 
         loginActivityViewModel.loggedIn.observe(this, Observer {
             Toast.makeText(context, "Success!", Toast.LENGTH_LONG).show()
-            redirectToMain()
+            loginActivityViewModel.fetchProfile()
         })
 
         rootView.email.addTextChangedListener(object : TextWatcher {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -78,6 +78,15 @@ class LoginFragment : Fragment() {
             loginActivityViewModel.sendResetPasswordEmail(email.text.toString())
         }
 
+        loginActivityViewModel.user.observe(this, Observer {
+            redirectToMain()
+        })
```
nikit19: Is this needed?

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -78,6 +78,15 @@ class LoginFragment : Fragment() {
...
+
+        loginActivityViewModel.loggedIn.observe(this, Observer {
+            Toast.makeText(context, "Success!", Toast.LENGTH_LONG).show()
+            loginActivityViewModel.fetchProfile()
+        })
```
nikit19: ` loginActivityViewModel.loggedIn.observe(this, Observer {
`
is already there don't add it again
add the above two lines there

dreadpool2: My bad! Will resolve it :+1:

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
dreadpool2: Yeah, actually, this will ensure that the user is redirected to the main page only after the user is fetched succesfully. Just as the user is fetched, it will then redirect to the main activity or else if I would have just used the `redirectToMain()`function in the loggedIn observer, the user would be redirected to the main page just after logging in, thus fetchProfile() won't be of any use :+1:

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
dreadpool2: PR Updated

PR #507, opened 2018-08-08
Alpha release

nikit19:


nikit19:
@iamareebjamal @simarsingh24 @dreadpool2 Please review

iamareebjamal:
I see a lot of commits being as long as Shakespeare story. Please follow commit guidelines and use commit description to write essays

Also, is the version code and name of the app incremented?

nikit19:
@iamareebjamal No but this is the first time we would be submitting the app to the playstore. Should we still increment it ?

iamareebjamal:
Yes, it is already on F-Droid

PR #514, opened 2018-08-09
fix: fix signing bug

### LINKED ISSUE #512: App doesn't work after signing
@iamareebjamal App doesn't work after signing it.
But works after I change the value of minifyEnabled
is this the right solution ?

nikit19:
fixes #512


dreadpool2:
Tested it locally, works fine :100:

PR #591, opened 2018-09-27
fix: Remove profile fragment after logout

### LINKED ISSUE #584: Profile details are shown on back pressed after logout
When user press back button after logout, profile details are shown due to the last fragment.

liveHarshit:
Fixes #584


liveHarshit:
Updated with development.

PR #595, opened 2018-09-29
Revamp Login Fragment

namanpreetsingh:
Login Fragment looks like this now ---
image
image
UI elements in Forgot Password were not well placed.

### DIFF
--- a/app/src/main/res/layout/fragment_login.xml
+++ b/app/src/main/res/layout/fragment_login.xml
@@ -82,9 +82,9 @@
                 android:id="@+id/forgotPassword"
                 android:layout_width="fill_parent"
                 android:gravity="center"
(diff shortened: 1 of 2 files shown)


dreadpool2:
This revamped version is not similar to EventBrite. I think it was quite fine before

dreadpool2:
@namanpreetsingh You did not raise an issue for this PR

dreadpool2:
@namanpreetsingh Yeah, but the check is present in the top most right corner in the latest version.
42723656_232939917580949_226431762471845888_n

namanpreetsingh:
@dreadpool2  But it doesn't quite look like it in my device. Why is it so?
screenshot_2018-09-29-22-26-29-426_com eventyay attendee

PR #615, opened 2018-10-05
fix: App exit on back pressed

### LINKED ISSUE #596: App exit when back button is pressed
### Issue

- When the back button is pressed in any fragment cause exit the app.
- App also exit when back button is pressed after search 

### Solution 
Like Eventbright - 

- Set event fragment when back button is pressed form another fragment.
- Close search view when back button is pressed after search

liveHarshit:
Fixes #596

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
@@ -6,6 +6,7 @@ import android.support.v4.app.Fragment
 import android.support.v7.app.AppCompatActivity
 import android.view.Menu
 import kotlinx.android.synthetic.main.activity_main.*
+import org.fossasia.openevent.general.R.id.navigation_events
 import org.fossasia.openevent.general.R.id.navigation_search
 import org.fossasia.openevent.general.attendees.AttendeeFragment
 import org.fossasia.openevent.general.auth.LAUNCH_ATTENDEE
@@ -110,5 +111,15 @@ class MainActivity : AppCompatActivity() {
         supportFragmentManager.beginTransaction()
                 .replace(R.id.frameContainer, fragment)
                 .commit()
+    }
+
+    override fun onBackPressed() {
+        val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
+        if (currentFragment !is EventsFragment) {
+            loadFragment(EventsFragment())
+            navigation.selectedItemId = navigation_events
+        } else {
+            super.onBackPressed()
+        }
     }
 }


app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -111,4 +112,31 @@ class MainActivity : AppCompatActivity() {
...
                 .commit()
     }
+
+    override fun onBackPressed() {
+        val currentFragmentClass = this.supportFragmentManager.findFragmentById(R.id.frameContainer).javaClass
+        when (currentFragmentClass) {
+            SearchFragment().javaClass -> {
+                loadFragment(EventsFragment())
+                supportActionBar?.title = "Events"
+                navigation.selectedItemId = navigation_events
```
simarsingh24: you are repeating the same code, think of some other way to acheive it

liveHarshit: Can I create a method for set event fragment?

simarsingh24: instead, you may use the logic `when currentFragmentClass is not EventFragment` since you want to get back to EventFragment in all case except current being EventFragment

liveHarshit: > instead, you may use the logic `when currentFragmentClass is not EventFragment` since you want to get back to EventFragment in all case except current being EventFragment

But if we want another condition in future like, need to load another fragment in a special case then it was a problem there. Should I change it?

simarsingh24: then it would be changed accordingly for now just change it :)

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -111,4 +112,21 @@ class MainActivity : AppCompatActivity() {
...
                 .commit()
     }
+
+    override fun onBackPressed() {
+        val currentFragmentClass = this.supportFragmentManager.findFragmentById(R.id.frameContainer).javaClass
+        when (currentFragmentClass) {
+            SearchFragment().javaClass -> setEventFragment()
+            FavoriteFragment().javaClass -> setEventFragment()
+            OrdersUnderUserFragment().javaClass -> setEventFragment()
+            ProfileFragment().javaClass -> setEventFragment()
```
simarsingh24: why not just do it when currentFragmentClass is not EventFragment, if in future we want some different functionality we can change it

liveHarshit: > why not just do it when currentFragmentClass is not EventFragment, if in future we want some different functionality we can change it

Updated.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -111,4 +112,13 @@ class MainActivity : AppCompatActivity() {
...
                 .commit()
     }
+
+    override fun onBackPressed() {
+        val currentFragmentClass = this.supportFragmentManager.findFragmentById(R.id.frameContainer).javaClass
+        if(currentFragmentClass != EventsFragment().javaClass) {
+            loadFragment(EventsFragment())
+            supportActionBar?.title = "Events"
+            navigation.selectedItemId = navigation_events
+        }
```
simarsingh24: else super on back pressed

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -111,4 +112,13 @@ class MainActivity : AppCompatActivity() {
                 .replace(R.id.frameContainer, fragment)
                 .commit()
     }
+
+    override fun onBackPressed() {
+        val currentFragmentClass = this.supportFragmentManager.findFragmentById(R.id.frameContainer).javaClass
+        if(currentFragmentClass != EventsFragment().javaClass) {
```
nikit19: space after if
Rest LGTM ! :+1:

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
liveHarshit: Oops, sorry for that. Let me update.

liveHarshit:
Updated.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -111,4 +112,15 @@ class MainActivity : AppCompatActivity() {
                 .replace(R.id.frameContainer, fragment)
                 .commit()
     }
+
+    override fun onBackPressed() {
+        val currentFragmentClass = this.supportFragmentManager.findFragmentById(R.id.frameContainer).javaClass
+        if (currentFragmentClass != EventsFragment().javaClass) {
```
iamareebjamal: Check using `instanceOf` 

Secondly, you are instantiating EventsFragment here - `EventsFragment().javaClass` Why?

iamareebjamal: My question is that why are you instantiating the fragment? Why are `()` after the class name?

Nevertheless, this won't be needed when you use `instanceOf`

iamareebjamal: Read about static fields of a class. However, implement this by `is` keyword of Kotlin

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -111,4 +112,15 @@ class MainActivity : AppCompatActivity() {
                 .replace(R.id.frameContainer, fragment)
                 .commit()
     }
+
+    override fun onBackPressed() {
+        val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
+        if (currentFragment !is EventsFragment) {
+            loadFragment(EventsFragment())
+            supportActionBar?.title = "Events"
```
iamareebjamal: This should not be needed

liveHarshit: What exactly, can't get it.

iamareebjamal: `supportActionBar?.title = "Events"`

PR #629, opened 2018-10-10
fix: App exit on back pressed

### LINKED ISSUE #626: You can't exit the app using the back button
**Describe the bug**
You can't exit the app using the back button
**To Reproduce**
Open the app and press the back button

- [ ] Yes
- [X] No
- Other:

GOVINDDIXIT:
fixes #626
### GIF for the change
ezgif com-video-to-gif

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
@@ -6,6 +6,7 @@ import android.support.v4.app.Fragment
 import android.support.v7.app.AppCompatActivity
 import android.view.Menu
 import kotlinx.android.synthetic.main.activity_main.*
+import org.fossasia.openevent.general.R.id.navigation_events
 import org.fossasia.openevent.general.R.id.navigation_search
 import org.fossasia.openevent.general.attendees.AttendeeFragment
 import org.fossasia.openevent.general.auth.LAUNCH_ATTENDEE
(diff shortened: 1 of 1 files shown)


GOVINDDIXIT:
@nikit19  pls review

GOVINDDIXIT:
Updated pls review

GOVINDDIXIT:
@iamareebjamal 
As the Fragment lifecycle do not have onBackPressed(). So According to me alternative method to override onBackPressed() in Android 3.0 fragments can be done by creating an interface class OnBackPressedListener and Create class which implements skills of OnBackPressedListener and creating a listener and a method to set listener. But the current changes are also preforming same job. So I goes with it.

iamareebjamal:
Fragments don't have onBackPressed, their container activity does. Find why it is not working

GOVINDDIXIT:
@nikit19 pls review

GOVINDDIXIT:
@iamareebjamal  The problem is with the stack generated. Currently the onBackPressed method is loading the event fragment  two times one with loadfragment and then with checkandLoad method. So I had cleared the stack generated everytime the user press the back button and located the user to Event Fragment which is required . Now the user can easily exit the as there is only 1 fragment in the stack.

GOVINDDIXIT:
@iamareebjamal  I think the issue is solved pls review

iamareebjamal:
This is still a workaround. I will see what can be done

iamareebjamal:
Old and conflicting with the current logic

PR #651, opened 2018-10-22
fix: Text overlapping in search fragment

### LINKED ISSUE #644: Text overlapping in search fragment
**Describe the bug**
Text overlapping(No search result and main content) when returning in search fragment from another after a search.

**To Reproduce**
Hide visibility of no search result content during returning from the search fragment.

**Expected behavior**
No overlapping of text.

**Screenshots**
<p align="center">
 
</p>

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Mi Max 2|
|Android Version|Nougat 7.1.1|

- [x] Yes

liveHarshit:
Fixes #644

Changes: Set search fragment without the saved state.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
@@ -88,6 +88,10 @@ class SearchFragment : Fragment() {
             it?.let { showNoInternetError(it) }
         })
 
+        searchViewModel.isSearched.observe(this, Observer {
+            it?.let { showSearchLayout(!it) }
+        })
+
         rootView.timeTextView.setOnClickListener {
             val intent = Intent(activity, SearchTimeActivity::class.java)
             startActivity(intent)
@@ -135,8 +139,6 @@ class SearchFragment : Fragment() {
             override fun onQueryTextSubmit(query: String): Boolean {
                 // Do your search
                 searchViewModel.searchEvent = query
-                rootView.searchLinearLayout.visibility = View.GONE
-                rootView.fabSearch.visibility = View.GONE
                 if (searchViewModel.savedLocation != null && TextUtils.isEmpty(rootView.locationTextView.text.toString()) && rootView.timeTextView.text == "Anytime")
                     searchViewModel.loadEvents(searchViewModel.savedLocation.nullToEmpty(), searchViewModel.savedDate.nullToEmpty())
                 else
@@ -157,6 +159,11 @@ class SearchFragment : Fragment() {
             queryListener.onQueryTextSubmit(searchView.query.toString())
         }
         super.onPrepareOptionsMenu(menu)
+    }
+
+    private fun showSearchLayout(show: Boolean) {
+        rootView.searchLinearLayout.visibility = if (show) View.VISIBLE else View.GONE
+        rootView.fabSearch.visibility = if (show) View.VISIBLE else View.GONE
     }
 
     fun handleVisibility(events: List<Event>) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -110,7 +110,7 @@ class MainActivity : AppCompatActivity() {
 
     private fun checkAndLoadFragment(fragment: Fragment) {
         val savedFragment = supportFragmentManager.findFragmentByTag(fragment::class.java.name)
-        if (savedFragment != null) {
+        if (savedFragment != null && fragment !is SearchFragment) {
```
iamareebjamal: This is not a fix. This is a workaround

liveHarshit: > This is not a fix. This is a workaround

But as in Eventbright everytime open search section, it opens with a new fragment so, I think no saved state is required for search fragment. That's why I direct load with new fragment if the search fragment is open. Should I fix this issue in another way?

iamareebjamal: Yes, the special case should be handled in a generic way while opening search fragment, not a special case in a method which should not even know about the existence of search fragment

liveHarshit:
@iamareebjamal Updated.

app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
```diff
@@ -147,6 +149,11 @@ class SearchFragment : Fragment() {
         super.onPrepareOptionsMenu(menu)
     }
 
+    private fun showSearchLayout(show: Boolean) {
+        rootView.searchLinearLayout.visibility = if (show) View.VISIBLE else View.GONE
+        rootView.fabSearch.visibility = if (show) View.VISIBLE else View.GONE
+    }
```
iamareebjamal: Again, this is a very special case workaround, not a fix. This will only invite newer bugs in future.

You should fix one function, the core logic and not add special ifs and elses regarding this particular fragment

trancenoid:
Are you still working on it ? @liveHarshit I will be happy to solve this issue

liveHarshit:
> Are you still working on it? @liveHarshit I will be happy to solve this issue

Yes.

simarsingh24:
@liveHarshit please finish this!

liveHarshit:
Updated. Now I check the saved state for search fragment. If yes, then hide the search layout for no overlapping with saved data. 
Please review.

iamareebjamal:
No changes in Activity are needed. Only changes should be in the fragment as it is the buggy component. Use fragment lifecycles or ViewModel to fix the bug

liveHarshit:
Updated.

app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
```diff
@@ -28,6 +28,7 @@ class SearchFragment : Fragment() {
     private lateinit var rootView: View
     private var loadEventsAgain = false
     private lateinit var searchView: SearchView
+    private var isSearched = false
```
iamareebjamal: Remove this variable from here, react to changes from viewmodel only

liveHarshit: Updated

app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
```diff
@@ -135,8 +139,7 @@ class SearchFragment : Fragment() {
             override fun onQueryTextSubmit(query: String): Boolean {
                 // Do your search
                 searchViewModel.searchEvent = query
-                rootView.searchLinearLayout.visibility = View.GONE
-                rootView.fabSearch.visibility = View.GONE
+                showSearchLayout(false)
```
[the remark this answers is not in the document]
liveHarshit: Updated.

iamareebjamal:
I request you to look at the first solution you submitted and this one and realize which one was elegant and maintainable. :smile: Good work

PR #687, opened 2018-11-05
fix: improve logic of onBackPressed for Fragments

### LINKED ISSUE #686: improve on back pressed logic to remove bugs
**Describe the bug**
- Last fragment is set on back pressed in event fragment
- Event fragment directly set from settings fragment

**To Reproduce**
1. One search/favorite/profile fragment
2. Press back again after the event fragment
3. Last fragment is set
4. Open settings fragment
5. Events are open instead of open profile fragment

**Expected behavior**
- All main fragment should set event fragment on back pressed
- Settings fragment should set profile fragment on back pressed

**Smartphone Info:**
|Device         |Mi Max 2|
|Android Version| Nougat 7.1    |

**Additional context**
blank diagram 3

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #686 

Changes: MainActivity.kt

blank diagram 3

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
@@ -128,11 +128,18 @@ class MainActivity : AppCompatActivity() {
     override fun onBackPressed() {
         val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
         val rootFragment = this.supportFragmentManager.findFragmentById(R.id.rootLayout)
-        if (currentFragment !is EventsFragment && rootFragment !is EventDetailsFragment) {
-            loadFragment(EventsFragment())
-            navigation.selectedItemId = navigation_events
-        } else {
+        if (rootFragment is EventDetailsFragment)
             super.onBackPressed()
-        }
+        else
+            when(currentFragment) {
+                is SearchFragment,
+                is FavoriteFragment,
+                is OrdersUnderUserFragment,
+                is ProfileFragment -> {
+                    loadFragment(EventsFragment())
+                    navigation.selectedItemId = navigation_events
+                }
+                else -> super.onBackPressed()
+            }
     }
 }


app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -126,13 +126,25 @@ class MainActivity : AppCompatActivity() {
...
-            navigation.selectedItemId = navigation_events
-        } else {
+        if (rootFragment is EventDetailsFragment) {
             super.onBackPressed()
+            return
         }
+
+        val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
+        when(currentFragment) {
+            is EventsFragment -> finish()
```
nikit19: This solution will work but the app should exit even without explicitly using this method.
I guess it is because the logic of `loadFragment` was changed. Earlier `addToBackStack` wasn't used in that method.

iamareebjamal:
First let us decide what the navigation flow should be. Can someone comment the expected navigation flow for the app

liveHarshit:
> First let us decide what the navigation flow should be. Can someone comment the expected navigation flow for the app

Okay, but here I follow same as Eventbrite :smile:

iamareebjamal:
Let's mention that here so other contributors from future can adhere to it

liveHarshit:
> Let's mention that here so other contributors from future can adhere to it

Added in PR description.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -121,18 +121,28 @@ class MainActivity : AppCompatActivity() {
     private fun loadFragment(fragment: Fragment) {
         supportFragmentManager.beginTransaction()
                 .replace(R.id.frameContainer, fragment, fragment::class.java.name)
-                .addToBackStack(null)
                 .commit()
     }
 
     override fun onBackPressed() {
-        val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
+
```
simarsingh24: extra feed

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -121,18 +121,28 @@ class MainActivity : AppCompatActivity() {
...
         val rootFragment = this.supportFragmentManager.findFragmentById(R.id.rootLayout)
-        if (currentFragment !is EventsFragment && rootFragment !is EventDetailsFragment) {
-            loadFragment(EventsFragment())
-            navigation.selectedItemId = navigation_events
-        } else {
+        if (rootFragment is EventDetailsFragment) {
             super.onBackPressed()
+            return
         }
+
```
simarsingh24: same

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -121,18 +121,28 @@ class MainActivity : AppCompatActivity() {
...
 
     override fun onBackPressed() {
-        val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
+
         val rootFragment = this.supportFragmentManager.findFragmentById(R.id.rootLayout)
-        if (currentFragment !is EventsFragment && rootFragment !is EventDetailsFragment) {
-            loadFragment(EventsFragment())
-            navigation.selectedItemId = navigation_events
-        } else {
+        if (rootFragment is EventDetailsFragment) {
```
simarsingh24: why is this needed if no case is true below super.onBackPressed will be executed. Please check and avoid repetitions

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
liveHarshit: Not extra... Here I separate methods for root container and frame container.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
liveHarshit: Same

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
iamareebjamal: No Kotlin style guide tells you to add a line feed at the start of method body

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
iamareebjamal: Because of the return

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -121,18 +121,28 @@ class MainActivity : AppCompatActivity() {
...
             super.onBackPressed()
+            return
```
iamareebjamal: Use else

liveHarshit:
Updated

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -121,18 +121,24 @@ class MainActivity : AppCompatActivity() {
     private fun loadFragment(fragment: Fragment) {
         supportFragmentManager.beginTransaction()
                 .replace(R.id.frameContainer, fragment, fragment::class.java.name)
-                .addToBackStack(null)
```
iamareebjamal: Why is this removed?

liveHarshit: With this, App will not exit with `super.onBackPressed()`. Need to use `finish()` explicitly for exit the app.

iamareebjamal: Without this, we don't have state saving of fragments

liveHarshit:
Updated. Now, one bug is there due to `addToBackStack(null)` which already has PR #629

simarsingh24:
@liveHarshit  please update on this pr's status

liveHarshit:
Due to my college examination, I'm not able to work on the PR. I'm really sorry about that. I will update the PR asap. And as I mentioned in my last comment, I it a right solution?

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -126,13 +126,20 @@ class MainActivity : AppCompatActivity() {
     }
 
     override fun onBackPressed() {
-        val currentFragment = this.supportFragmentManager.findFragmentById(R.id.frameContainer)
```
simarsingh24: why is this removed and then written again? Rest LGTM 👍

liveHarshit: First I declare root fragment then current fragment in sequence.

liveHarshit: Updated.

PR #692, opened 2018-11-13
fix: Fix the no connectivity case bug on Main Activity

### LINKED ISSUE #691: No internet case not handled properly in the main activity
**Description**

Once the events on the main screen have been loaded, if the internet connection is lost, and the user refreshes the list, the code makes the request to the server, hence the SwipeRefreshLayout keeps refreshing for a while (about 8-10 secs), after which a toast appears: "Error fetching events"

**To Reproduce**

1. Go to main activity with the internet connection working, let it load all events.
2. Turn on flight mode.
3. Refresh the events list.

**Expected behavior**

The no internet card should appear instantaneously, the app should not initiate call to the server (via retryLoadLocationEvents() method), and the swipe refresh layout should stop refreshing.

**Screenshots**

screenshot_1542134680

Yes, working on it.

aleronarjun:
Improve logic of code to handle no connectivity better on refresh by showing the no internet card instantaneously, hide the refreshing layout, and not initiating non-required calls to network methods.

Fixes #691 
Changes: 

The code logic has been modified to handle the issue mentioned.

Screenshots for the change:
Before:
Refreshing the event list (one fully loaded) under no internet:
(This state remains for a long time as the connection methods are called, and finally terminates after 8-10 secs with a toast)

screenshot_1542134680

After making the changes, and refreshing the events list (once fully loaded) under no internet:

screenshot_1542135460

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
@@ -121,7 +121,12 @@ class EventsFragment : Fragment() {
 
         rootView.swiperefresh.setColorSchemeColors(Color.BLUE)
         rootView.swiperefresh.setOnRefreshListener {
-            eventsViewModel.retryLoadLocationEvents()
+            showNoInternetScreen(isNetworkConnected())
+            if (!isNetworkConnected()) {
+                rootView.swiperefresh.isRefreshing = false
+            } else {
+                eventsViewModel.retryLoadLocationEvents()
+            }
         }
 
         return rootView
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,5 @@ script:
 after_success:
     - bash scripts/prep-key.sh
     - bash scripts/update-apk.sh
+before_install:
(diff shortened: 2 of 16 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -121,7 +121,13 @@ class EventsFragment : Fragment() {
...
         rootView.swiperefresh.setColorSchemeColors(Color.BLUE)
         rootView.swiperefresh.setOnRefreshListener {
-            eventsViewModel.retryLoadLocationEvents()
+            showNoInternetScreen(isNetworkConnected())
+            if(!isNetworkConnected()){
+                rootView.swiperefresh.isRefreshing = false
+            }
+            else{
+                eventsViewModel.retryLoadLocationEvents()
+            }
```
iamareebjamal: Not formatted correctly

aleronarjun:
@nikit19 Hi, I squashed all commits. Please review again. Thank you so much.

iamareebjamal:
You squashed **all** the previous commits, instead of just yours

PR #717, opened 2018-12-08
fix: Remove continuous toggling between fragments

### LINKED ISSUE #716: App keeps toggling between Events and Profile fragment
**Describe the bug**
After going back from Profile fragment to Events, if we press back button again then Profile fragment opens up (although on the navigation bar, 'Events' remains selected) and on again pressing back, it goes to Events. This process can be done again and again.

**To Reproduce**
1. Go to Profile.
2. Click on back button.
3. Events will open.
4. Again go back.
5. Now, profile will open.
6. Again go back.
7. Again events will open.
8. This will continue.

**Expected behavior**
On pressing back from the Events screen, app should close.

**GIF**
ezgif com-video-to-gif 11

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #716 

Changes: 
Instead of toggling between fragments, the app will now close when back is pressed on the Events fragment.

GIF for the change:

ezgif com-video-to-gif 12

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
@@ -139,6 +139,10 @@ class MainActivity : AppCompatActivity() {
                     loadFragment(EventsFragment())
                     navigation.selectedItemId = navigation_events
                 }
+                is EventsFragment -> {
+                    android.os.Process.killProcess(android.os.Process.myPid());
+                    System.exit(1);
+                }
                 else -> super.onBackPressed()
             }
     }


ShridharGoel:
@iamareebjamal Fixed.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -121,7 +121,6 @@ class MainActivity : AppCompatActivity() {
     private fun loadFragment(fragment: Fragment) {
         supportFragmentManager.beginTransaction()
                 .replace(R.id.frameContainer, fragment, fragment::class.java.name)
-                .addToBackStack(null)
```
iamareebjamal: This is required to save fragments and their data between switching tabs and other things

ShridharGoel: Used finish() so that the app closes when back is pressed while on the Events fragment. Is it okay now?

PR #734, opened 2018-12-12
fix: the the back-button functionality in Ticket-Fragment.

### LINKED ISSUE #726: back stack issue in ticket details
**Describe the bug**

app goes in background if you press the back button in ticket details or back option on top menu , back stack also produces problem when you access ticket detail through likes window(Likes->selecta ny event->tickets->press back button)

**To Reproduce**
1. Select any event 
2. Select Tickets
3. Select BackArrow on top Menu(or press back button)

In likes pannel
1.go to likes pannel
2. Select any event 
3. Select Tickets
4. Select BackArrow on top Menu(or press back button)

**Expected behavior**
should go back to event details pannel from which we have selected tickets option

**Smartphone Info:**

|Device    ->redmi note 4
|Android Version| ->7.0

- [ x] Yes
- [ ] No

pranavpandey1998official:
Fixes #726 

Changes: the issue arises because eventDetailFragment when the transition to TicketFragment replaces R.id.root fragment which is not present in the back stack(as was replaced previously when transitioning to eventDetailFragment), it has to replace the EventDetailFragment, so I give eventDetailFragment id= R.id.coordinatorLayout and replaces that in fragment transition from eventDetailFragment to TicketFragment
Also 'replace' method is used in place of 'add' in the transition from favoriteFragment to EventDetailFragment so onCreatview of favoriteFragment is called when coming back to favoriteFragment

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -204,7 +204,6 @@ class EventDetailsFragment : Fragment() {
         super.onDestroyView()
     }
 
-
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         return when (item.itemId) {
             android.R.id.home -> {
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -284,7 +284,6 @@ class EventDetailsFragment : Fragment() {
         menuActionBar = menu
     }
 
-
     private fun loadTicketFragment() {
         // Initialise Ticket Fragment
         val ticketFragment = TicketsFragment()
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -248,7 +248,7 @@ class EventDetailsFragment : Fragment() {
         }
     }
 
-    override fun onPrepareOptionsMenu(menu: Menu?) 
+    override fun onPrepareOptionsMenu(menu: Menu?) {
         menu?.setGroupVisible(R.id.search_menu, false)
         menu?.setGroupVisible(R.id.event_menu, true)
         super.onPrepareOptionsMenu(menu)
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -248,10 +248,7 @@ class EventDetailsFragment : Fragment() {
         }
     }
 
-    override fun onPrepareOptionsMenu(menu: Menu?) {
-        val activity = activity as? MainActivity
-        activity?.supportActionBar?.setDisplayHomeAsUpEnabled(true)
-        activity?.supportActionBar?.title = ""
+    override fun onPrepareOptionsMenu(menu: Menu?) 
         menu?.setGroupVisible(R.id.search_menu, false)
         menu?.setGroupVisible(R.id.event_menu, true)
         super.onPrepareOptionsMenu(menu)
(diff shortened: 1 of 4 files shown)


iamareebjamal:
Keep PR title short and use the description field for the story. :+1: Will review shortly

iamareebjamal:
Meanwhile rename PR to fulfill semantic PR guidelines

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -203,6 +204,7 @@ class EventDetailsFragment : Fragment() {
         super.onDestroyView()
     }
 
+
```
nikit19: Not required

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -280,14 +285,15 @@ class EventDetailsFragment : Fragment() {
         menuActionBar = menu
     }
 
+
```
nikit19: Not required

app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
```diff
@@ -91,4 +92,8 @@ class FavoriteFragment : Fragment() {
         rootView.favoriteProgressBar.isIndeterminate = show
         rootView.favoriteProgressBar.visibility = if (show) View.VISIBLE else View.GONE
     }
+
+
+
+
```
nikit19: Not required

app/src/main/java/org/fossasia/openevent/general/ticket/TicketsFragment.kt
```diff
@@ -144,7 +145,9 @@ class TicketsFragment : Fragment() {
     override fun onDestroyView() {
         val activity = activity as? MainActivity
         activity?.supportActionBar?.setDisplayHomeAsUpEnabled(false)
+        invalidateOptionsMenu(activity)
         super.onDestroyView()
+
```
nikit19: Not required

pranavpandey1998official:
@simarsingh24 I have resolved the respective issue can you please review it

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -248,6 +249,9 @@ class EventDetailsFragment : Fragment() {
     }
 
     override fun onPrepareOptionsMenu(menu: Menu?) {
+        val activity = activity as? MainActivity
+        activity?.supportActionBar?.setDisplayHomeAsUpEnabled(true)
+        activity?.supportActionBar?.title = ""
```
iamareebjamal: Why here?

pranavpandey1998official: added as a fail check ,as the ActionBar is changing when going from ticketFragment to Event detail fragment so i invalidated the optionmenu in ticketFragment and set the changing login in onPrepareOptionMenu

iamareebjamal: But it is not an option menu and invalidating option menu should not change action bar.

app/src/main/java/org/fossasia/openevent/general/ticket/TicketsFragment.kt
simarsingh24: @pranavpandey1998official why is this marked resolved?

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
pranavpandey1998official: K. I will remove it. but something has to be done, as some issue with the action bar remains in the code , you can see it when going from favoritefragment to eventdetailfragment and back the title becomes events instead of likes I will open a new issue regarding this, as I can see the problem is that the removed fragment is changing the action bar title instead of the one that is going to be displayed.

liveHarshit:
@pranavpandey1998official Are you still working on it? Please handle the issue from MainActivity method. Because the problem is in the onBackPressed method only.

Issue #738, opened 2018-12-16
Rotating screen anywhere opens Tickets fragment


ShridharGoel:
**Describe the bug**

Rotating the screen on any fragment opens up the Tickets fragment, even though the selected item on the navigation bar remains the same and does not change to Tickets.

**To Reproduce**

1. Go to any fragment.
2. Turn on screen rotation if it is off.
3. Rotate the mobile.
4. App will switch to Tickets fragment.

**Expected behavior**
The fragment should not have changed. 

**GIF**

![ezgif com-video-to-gif 4](https://user-images.githubusercontent.com/35566748/50050957-5d7b5d00-012e-11e9-9448-7040bc8346ad.gif)

**Would you like to work on the issue?**
- [x] Yes
- [ ] No

PR #748, opened 2018-12-19
fix: Handle configuration changes for MainActivity

### LINKED ISSUE #744: Event fragment is set during selecting the search parameter
**Describe the bug**
Search perform in events after entering a location in the search section

**To Reproduce**
1. Go to the search section
2. Try to enter a location
3. Event fragment is open instead of changing the location

**Screenshots**
<p align="center">
 
</p>

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Mi Max 2|
|Android Version|Nougat 7.1|

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #744 #738

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
@@ -69,26 +69,21 @@ class MainActivity : AppCompatActivity() {
 
         supportActionBar?.title = "Events"
 
-        val bundle = intent.extras
-
-        if (savedInstanceState == null) {
-            loadFragment(supportFragmentManager, EventsFragment(), frameContainer.id)
-        }
-
-        else {
-            if (bundle != null && bundle.getBoolean(TO_SEARCH)) {
+        val bundle = if (savedInstanceState == null) intent.extras else null
+        if (bundle != null) {
+            if (bundle.getBoolean(TO_SEARCH)) {
                 loadFragment(supportFragmentManager, SearchFragment(), frameContainer.id)
                 supportActionBar?.title = "Search"
                 navigation.selectedItemId = navigation_search
             }
 
-            if (bundle != null && bundle.getBoolean(LAUNCH_ATTENDEE)) {
+            if (bundle.getBoolean(LAUNCH_ATTENDEE)) {
                 val fragment = AttendeeFragment()
                 fragment.arguments = bundle
                 loadFragment(supportFragmentManager, fragment, frameContainer.id)
             }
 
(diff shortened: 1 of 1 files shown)


liveHarshit:
Changes in PR #739 contains bugs because, in many places, Flag activity clear top is used. In that case, while intent to the main activity

if (savedInstanceState == null) {
            loadFragment(supportFragmentManager, EventsFragment(), frameContainer.id)
        }

        else
Only if statement is executed and all arguments are not used.

liveHarshit:
When device orientation is changed, activity receives the previous bundle, due to this Ticket fragment is set on rotation if authentication done by ticket fragment.

Here I set null value to the bundle if the orientation is changed, otherwise accept bundle from the intent.

liveHarshit:
@iamareebjamal @nikit19 Please review.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -69,30 +69,24 @@ class MainActivity : AppCompatActivity() {
 
         supportActionBar?.title = "Events"
 
-        val bundle = intent.extras
+        val bundle = if (savedInstanceState == null) intent.extras else null
 
-        if (savedInstanceState == null) {
-            loadFragment(supportFragmentManager, EventsFragment(), frameContainer.id)
```
iamareebjamal: How does the new code handle this case?

liveHarshit: How does this case is used?, I don't understand. Because both issues are resolved with set bundle to null when orientation changed.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -69,30 +69,24 @@ class MainActivity : AppCompatActivity() {
...
 
-            if (bundle != null && (bundle.getBoolean(TICKETS) || bundle.getBoolean(LAUNCH_TICKETS))) {
-                loadFragment(supportFragmentManager, OrdersUnderUserFragment(), frameContainer.id)
-                supportActionBar?.title = "Tickets"
-                navigation.selectedItemId = R.id.navigation_tickets
-            }
+        if (bundle != null && (bundle.getBoolean(TICKETS) || bundle.getBoolean(LAUNCH_TICKETS))) {
+            loadFragment(supportFragmentManager, OrdersUnderUserFragment(), frameContainer.id)
+            supportActionBar?.title = "Tickets"
+            navigation.selectedItemId = R.id.navigation_tickets
```
iamareebjamal: MOove all 3 cases in a block

```
if (bundle != null) {
   ...
}
```

PR #761, opened 2018-12-24
fix: Set event fragment at the start

### LINKED ISSUE #760: No fragment is set at the start
**Describe the bug**
Event fragment is not set at the start.
Logic was removed in PR #739 and PR #748 by mistake.

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #760


app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -68,6 +68,7 @@ class MainActivity : AppCompatActivity() {
         navigation.setOnNavigationItemSelectedListener(listener)
 
         supportActionBar?.title = "Events"
+        loadFragment(supportFragmentManager, EventsFragment(), frameContainer.id)
```
iamareebjamal: This is unconditionally loading it.

liveHarshit: Can I move it in onCreate(), because it is called only once when the app starts?

iamareebjamal: There are if conditions below, add it there

PR #768, opened 2018-12-25
fix: Shift ProgressBar in Tickets to middle of screen

### LINKED ISSUE #757: While loading tickets progress bar should be in the middle of the screen
**Describe the bug**

The progress is at the top of the screen which seems a bit odd since the whole screen is blank at that time. It should be placed in the middle.

**To Reproduce**
1. Go to Tickets fragment.
2. Progress bar appears at the top.

**Expected behavior**

Progress bar should have appeared at the middle of the screen.

**Screenshot**

img_20181223_002731

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #757 

Changes: 

Now ProgressBar will show in the middle of the screen instead of the top in Tickets.

Screenshots for the change:

screenshot_2018-12-25-19-27-56-533_com eventyay attendee

### DIFF
--- a/app/src/main/res/layout/fragment_orders_under_user.xml
+++ b/app/src/main/res/layout/fragment_orders_under_user.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/fragment_orders_under_user.xml
```diff
@@ -1,33 +1,40 @@
-<androidx.core.widget.NestedScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
iamareebjamal: CoordinatorLayout

ShridharGoel: Done

Issue #807, opened 2019-01-02
miss alignment of share image in the fragment order


pranavpandey1998official:
**Describe the bug**
miss alignment of share image in the fragment order

**To Reproduce**
1.do a successful order and see the share button in order complete

**Screenshots**
![whatsapp image 2019-01-02 at 1 17 41 pm](https://user-images.githubusercontent.com/44601530/50583716-f16c5900-0e90-11e9-8e06-633618b15625.jpeg)

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Redmi note 4|
|Android Version|7.0  |

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Set scale type programmatically, it is sdk 28 error.

PR #827, opened 2019-01-04
fix: Misalignment of share image in order fragment

pranavpandey1998official:
programitically setting the scale type of share button of fragment_order fragment

fixes: #807


iamareebjamal:
The alignment was not missing and there was no problem in fragment order

Issue #818, opened 2019-01-03
Ask location from the user for first time


liveHarshit:
**Describe the feature you'd like**
Add a welcome screen for ask location from the user.

**Screenshots**
<p align="center">
<img src="https://user-images.githubusercontent.com/27783786/50626559-f0046480-0f54-11e9-8d79-3085fa01cd32.png" height = "480" width="270"> 
</p>

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

iamareebjamal:
/remind me to check after 1 day

PR #855, opened 2019-01-06
feat: Configure location feature with MVVM

### LINKED ISSUE #817: Events fragment is open after selecting location from search
**Describe the bug**
Event fragment is set after selecting the location from search due to intent. 

- [x] Yes

### LINKED ISSUE #867: Polish location feature implementation
Currently, location feature is configured to use only from search location fragment. 

- [x] Yes

### LINKED ISSUE #875: Navigate Back Button in SearchLocationFragment doesn't works
**Navigate Back Button in SearchLocationFragment doesn't works**
Steps To reproduce:
- Go to the search fragment
- Tap the text where
- In the next activity press the navigate up button in toolbar

Issue:
The navigate back button should have taken you to the previous activity but it didn't

- Yes

liveHarshit:
Fixes #867 
Fixes #875


iamareebjamal:
OK, so there are some changes which should be there in its implementation. I'll outline them tomorrow

liveHarshit:
There are too many conflicts in this PR why I try to remove search location activity. It is a minor issue. Can I solve both in one PR? 
And for polish the location implementation this PR can be used.

iamareebjamal:
OK

iamareebjamal:
The configure method does too much, it should take required arguments in constructor and just use methods to do actions from activity

liveHarshit:
> just use methods to do actions from activity

Mean call for location request from the main activity?

liveHarshit:
In case of issue #818, location is asked from the user at the start.

iamareebjamal:
Yes, but it is still an action and may be placed somewhere else in the future.

liveHarshit:
Updated

liveHarshit:
@iamareebjamal Updated. Do review.

iamareebjamal:
OK, so we should not need build flavor checking at all. They are already in different build flavors, so no checking that

liveHarshit:
> OK, so we should not need build flavor checking at all. They are already in different build flavors, so no checking that

Then how to handle visibility for the current location option? Because it should not visible for fdroid.

iamareebjamal:
You have made it a view model, so you create a live data called current location visibility and just push true on it in case of fdroid vm and do nothing in case of google play vm

If it wasn't a vm, you would just pass the views in a function called initalizeView, which would be no op for google play, but would hide in fdroid

liveHarshit:
> You have made it a view model, so you create a live data called current location visibility and just push false on it in case of fdroid vm and do nothing in case of google play vm

Great idea. :raised_hands:

liveHarshit:
Updated

app/src/fdroid/java/org/fossasia/openevent/general/search/GeoLocationViewModel.kt
```diff
@@ -0,0 +1,21 @@
...
+
+class GeoLocationViewModel : ViewModel() {
+    private val mutableLocation = MutableLiveData<String>()
+    val location: LiveData<String> = mutableLocation
+    private val mutableVisibility = MutableLiveData<Boolean>()
+    val currentLocationVisibility: LiveData<Boolean> = mutableVisibility
+
+    fun getVisibility() {
+        mutableVisibility.value = false
+    }
```
iamareebjamal: No need of get visibility. Just do this in configure

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -35,26 +37,29 @@ class SearchLocationFragment : Fragment() {
...
         }
-        rootView.locationProgressBar.visibility = View.GONE
+        setHasOptionsMenu(true)
 
         fromSearchFragment = arguments?.getBoolean(FROM_SEARCH) ?: false
 
-        if (thisActivity is Activity) geoLocationUI.configure(thisActivity, rootView, searchLocationViewModel)
+        geoLocationViewModel.getVisibility()
+        geoLocationViewModel.currentLocationVisibility.observe(this, Observer {
+            currentLocation.visibility = View.GONE
```
iamareebjamal: Set it to it

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -35,26 +37,29 @@ class SearchLocationFragment : Fragment() {
...
+        })
+
+        rootView.currentLocation.setOnClickListener {
+            checkLocationPermission()
+            if (thisActivity is Activity) geoLocationViewModel.configure(thisActivity)
```
iamareebjamal: When will thisActivity not be an activity?

app/src/fdroid/java/org/fossasia/openevent/general/search/GeoLocationViewModel.kt
liveHarshit: But location request is from the fragment. Configure method is required location permission if I call it at the start. And location permission cannot be called at the start because it is not required for fdroid. So, I call configure conditionally.

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
liveHarshit: thisActivity is activity inside on create view.

app/src/fdroid/java/org/fossasia/openevent/general/search/GeoLocationViewModel.kt
iamareebjamal: OK, so just call this in onStart

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
iamareebjamal: Why is the if statement required

liveHarshit: But activity is from getActivity() of type FragmentActivity, need to cast for Activity

liveHarshit: > When will thisActivity not be an activity?

Okay got it. It is Activity for following if statement: 
```java
val thisActivity = activity
        if (thisActivity is AppCompatActivity) {
            thisActivity.supportActionBar?.title = ""
            thisActivity.supportActionBar?.setDisplayHomeAsUpEnabled(true)
        }
```

app/src/fdroid/java/org/fossasia/openevent/general/search/GeoLocationViewModel.kt
liveHarshit: Updated

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
iamareebjamal: It doesn't need to be smart casted, just pass it directly

liveHarshit: It is showing required: Activity found: FragmentActivity

app/src/fdroid/java/org/fossasia/openevent/general/search/GeoLocationViewModel.kt
iamareebjamal: Still same

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
iamareebjamal: FragmentActivity is child of Activity, so there shouldn't be a problem

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -76,16 +81,42 @@ class SearchLocationFragment : Fragment() {
...
 
-    @SuppressLint("MissingPermission")
+    override fun onStart() {
+        super.onStart()
+        geoLocationViewModel.getVisibility()
+    }
+
+    private fun checkLocationPermission() {
+        val permission = context?.let {
+            ContextCompat.checkSelfPermission(it, Manifest.permission.ACCESS_COARSE_LOCATION) }
```
iamareebjamal: Remove `context.let` for requireContext()

app/src/fdroid/java/org/fossasia/openevent/general/search/GeoLocationViewModel.kt
liveHarshit: I have called it from on start of the fragment. I am sorry to not getting it.

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
iamareebjamal: Make it nullable in the receiving function and just return if it is null

liveHarshit:
Updated

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -37,26 +40,30 @@ class SearchLocationFragment : Fragment() {
...
+        geoLocationViewModel.configure(thisActivity)
+
+        rootView.currentLocation.setOnClickListener {
+            checkLocationPermission()
+            geoLocationViewModel.configure(thisActivity)
+            rootView.locationProgressBar.visibility = View.VISIBLE
+            geoLocationViewModel.location.observe(this, Observer { location ->
+                searchLocationViewModel.saveSearch(location)
+                redirectToMain()
+            })
```
iamareebjamal: Move outside of click listener

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -74,16 +81,37 @@ class SearchLocationFragment : Fragment() {
...
+
     override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<String>, grantResults: IntArray) {
         when (requestCode) {
             LOCATION_PERMISSION_REQUEST -> {
                 if (grantResults.isNotEmpty() && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
                     val thisActivity = activity
-                    if (thisActivity is Activity)
-                        geoLocationUI.configure(thisActivity, rootView, searchLocationViewModel)
-                } else
+                    if (thisActivity is Activity) geoLocationViewModel.configure(thisActivity)
```
iamareebjamal: Directly pass activity

liveHarshit: Sorry, it is by mistake.

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
liveHarshit: if I move redirect to main outside click listener with the observer, After opening location fragment again, value for current location fragment is already assigned. And redirectToMain() called instantly. So, user not able to change the location again.

iamareebjamal: Then make the location a single event live data

liveHarshit: Same result with `SingeLiveEvent`. I think because of configure method is called at start. And location permission is already granted.

liveHarshit:
Updated

Issue #857, opened 2019-01-07
add no event found text to eventfragment


pranavpandey1998official:
**Describe the feature you'd like**
right now if some place dont have any event then the eventFragment just displayes a black screen which is not a good user experience it should display text saying "Oops ,looks like this area don't have any registered events" 

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:
your opinion
@iamareebjamal
@nikit19

shiv07tiwari:
@pranavpandey1998official Can I take up this issue please ?

theLimitBreaker:
This is implemented as of current . @nikit19 @iamareebjamal  I think its irrelevant now

PR #868, opened 2019-01-10
feat: added no search result found text to event fragment

pranavpandey1998official:
added showNoEventText dunction and called it inside evenetViewModel.events observer when list is empty
added noEventLL to fragment_event

ScreenShot
device-2019-01-10-044252

feat: #857


app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -147,6 +149,11 @@ class EventsFragment : Fragment() {
         return rootView
     }
 
+    private fun showNoEventText(show: Boolean) {
```
nikit19: `showNoEventText` you are also showing an image. This is a bit confusing

pranavpandey1998official: should I name it to showNoEventPannel or showNoEventLayout

RaghavAwasthi:
@pranavpandey1998official You should squash your commits

pranavpandey1998official:
@iamareebjamal please review it

addiegupta:
> @iamareebjamal please review it

Squash your commits. One PR should contain 1 commit

Issue #785, opened 2018-12-28
Remove toasts in favor of Snackbars


iamareebjamal:
Toasts are kind of deprecated as any user who has disabled notifications for an app won't be able to see toast as well

Change toasts to snackbars wherever used

GOVINDDIXIT:
I want to take this issue

iamareebjamal:
Go ahead

liveHarshit:
@GOVINDDIXIT After login, snack for Welcome back! is shown in the login fragment. But the fragment is instantly pop out. Snack after auth should be shown in redirected fragment.

liveHarshit:
@GOVINDDIXIT SnackBar is displaying at the top.
<p align="center">
<img src="https://user-images.githubusercontent.com/27783786/51269866-8d49a900-19e9-11e9-9d40-1d8c7bca9859.png" height = "480" width="270"> 
</p>

PR #879, opened 2019-01-12
fix: toast to snackbar for events related fragments

GOVINDDIXIT:
part of #785


GOVINDDIXIT:
@iamareebjamal  pls review

PR #838, opened 2019-01-05
fix: remove post handlers

GOVINDDIXIT:
Fixes #830 
fixes #840 

### Description
Remove posthandler that are defering the transition and improve the logic to show snackbar

GIF for the change:

#### Signup

gif-190105_121407

#### login
ezgif com-video-to-gif 1

PR #886, opened 2019-01-13
fix: toast to snackbar for search result and order related fragments

GOVINDDIXIT:
part of #785


GOVINDDIXIT:
@iamareebjamal  pls review 
also pls review #838 so that #785 can be completely solved

PR #921, opened 2019-01-18
fix: Snackbar displaying message twice in login fragment

### LINKED ISSUE #920: Remove snackbar in login fragment
**Describe the bug**
 when user clicks on profile button in bottom navigationbar the login fragment appears along with snackbar having message "You need to log in first!". snackbar is being called twice to display the information. But I think that snackbar is redundant as the user is there to login to his account. snackbar with this information would be helpful in places like user accessing information which requires login. 

**To Reproduce**

1. Go to homeActivity
2. Click on profile
3. snackbar appears with the message

**Expected behavior**

No need of snackbar with message "You need to log in first!" when user is already there for login.

**Screenshots**

ezgif-5-42e577731998

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | [ Redmi Note 5] |
|Android Version|[Oreo 8.1]  |

- [x] Yes

Cool-fire:
Fixes #920 

Screenshots for the change:
screenshot_2019-01-18-22-55-29-205_com eventyay attendee

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
@@ -52,7 +52,7 @@ class LoginFragment : Fragment() {
         setHasOptionsMenu(true)
         showSnackbar()
 
-        showSnackbar()
+
 
         if (loginViewModel.isLoggedIn())
             redirectToMain()
--- a/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
@@ -52,7 +52,6 @@ class LoginFragment : Fragment() {
         setHasOptionsMenu(true)
         showSnackbar()
 
-
         if (loginViewModel.isLoggedIn())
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/LoginViewModel.kt
```diff
@@ -37,7 +37,8 @@ class LoginViewModel(
 
     fun login(email: String, password: String) {
         if (!isConnected()) return
-        if (hasErrors(email, password)) return
+        if (hasErrors(email, password))
+            return
```
liveHarshit: Unnecessary changes.

Cool-fire: I did that for debugging purpose, can please suggest how to revert, do you want me to add a commit with neccesary changes and then submit the pull request.

liveHarshit: Yes, do changes related to issue only. It is an extra committed file which is not related to the issue. You can drop the commit, revert changes and push again to your local branch. PR will update automatically.

Issue #922, opened 2019-01-19
Press back twice will exit from fragment


liveHarshit:
**Describe the bug**
For search location fragment, press back button twice will exit from the fragment. 

**To Reproduce**
1. Go to Location
2. Press back

**Expected behavior**
It should exit in one press.

**Would you like to work on the issue?**

- [x] No
- Other: Open for all

yashk2000:
@liveHarshit can I work on this?

liveHarshit:
> @liveHarshit can I work on this?

Go ahead.

yashk2000:
@liveHarshit when search location fragment is opened, shouldn't the keyboard come up on the screen automatically?

angmas1:
This is because a keyboard is opened but isn't visible. 1st back cancels keyboard 2nd back exits fragment.

angmas1:
> This is because a keyboard is opened but isn't visible. 1st back cancels keyboard 2nd back exits fragment.

Why is this issue still open? This issue was resolved by the recent pull request. It wasn't exiting because an invisible keyboard was open.

liveHarshit:
@angmas1 You should mention Fixes for this issue in PR.

angmas1:
I did mention, but not in correct format, so it didn't close. I am sorry.

PR #929, opened 2019-01-19
fix: Keyboard shows up automatically when search location fragment opens.

### LINKED ISSUE #928: Keyboard does not show up automatically in search location fragment.
**Describe the bug**

When the search location fragment opens, the keyboard does not show up automatically.

**To Reproduce**
1. Goto search location fragment.
2. The keyboard does not automatically. The user has to press "where" to open the keyboard.

**Expected behavior**

The keyboard should open automatically so that user can start typing right away. This will also fix #922.

- [X] Yes
- [ ] No

yashk2000:
Fixes #928, #922 

Changes: Keyboard pops up automatically when the search location fragment is opened. Hence the user can start typing straight away instead of first tapping "where". Now if the back press is tapped once, the keyboard goes down and when the back press is tapped again, the user is sent back to the main activity.

GIF for the change:

ezgif com-video-to-gif 5

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
@@ -34,6 +34,8 @@ class SearchLocationFragment : Fragment() {
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         rootView = inflater.inflate(R.layout.fragment_search_location, container, false)
 
+        Utils.showSoftKeyboard(context, view)
+
         val thisActivity = activity
         if (thisActivity is AppCompatActivity) {
             thisActivity.supportActionBar?.show()
--- a/app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
@@ -44,9 +44,9 @@ object Utils {
             .show()
     }
 
-    fun showSoftKeyboard(context: Context?, view: View) {
+    fun showSoftKeyboard(context: Context?, view: View?) {
         val manager = context?.getSystemService(Context.INPUT_METHOD_SERVICE)
-        if (manager is InputMethodManager) manager.showSoftInput(view, InputMethodManager.SHOW_FORCED)
+        if (manager is InputMethodManager) manager.showSoftInput(view, InputMethodManager.RESULT_UNCHANGED_SHOWN)
     }
 
     fun hideSoftKeyboard(context: Context?, view: View) {


app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -34,6 +36,9 @@ class SearchLocationFragment : Fragment() {
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         rootView = inflater.inflate(R.layout.fragment_search_location, container, false)
 
+        val imm = context?.getSystemService(Context.INPUT_METHOD_SERVICE) as? InputMethodManager
+        imm?.showSoftInput(view, 0)
```
liveHarshit: Same method is written in utils. You can call it from there. :+1:

yashk2000: @liveHarshit I tried doing it that way, but then the keyboard did not show up at all.

liveHarshit: Okay, let me check. 
What is view here, pass it. And change `InputMethodManager.SHOW_FORCED` to `InputMethodManager.RESULT_UNCHANGED_SHOWN` in utils which contains constant value 0. Then it should work.

yashk2000: @liveHarshit I replaced lines 39 and 40 in searchLocationFragment with `Utils.showSoftKeyboard(context, rootView)`. In Utils also I made the change that you mentioned above. But the keyboard still does not pop up.

liveHarshit: `imm?.showSoftInput(view, 0)`
What is view here?

yashk2000: That is View?, not view. It is from getView().

liveHarshit: Ok then change view to nullable in utils, same as context.

yashk2000: It works. I have not made any changes in the onResume() function. It is still `Utils.showSoftKeyboard(context, rootview)`

liveHarshit: But remove it! It was written for the same purpose earlier.

yashk2000: I removed it, it works fine. But now when the user presses on the back button on the top of the screen, the keyboard does not go down. It stays up in the main activity and back button on the nav bar has to pressed to close the keyboard.

yashk2000: @liveHarshit Should I revert back to the initial commit I made? That way the keyboard was going down automatically even when the user went to the main activity by pressing the back button in the top.

liveHarshit: > And change `InputMethodManager.SHOW_FORCED` to `InputMethodManager.RESULT_UNCHANGED_SHOWN` in utils which contains constant value 0. Then it should work.

You forgot these change. Do it and your initial changes and these changes become same,

yashk2000:
@liveHarshit it works now. Thanks for the help!

PR #955, opened 2019-01-25
fix: Show error in edittext when email is invalid

### LINKED ISSUE #927: No error message when user clicks on login button with email and password fields empty.
**Describe the bug**

In the login fragment when the user clicks on login button without entering email and password fields nothing happens, I think there should be error message like "email and password fields are required for login". I think this error message is required because it will enhance user experience.

**To Reproduce**
1. Go to loginFragment
2. Click on login button without entering email, password fields.
3. No error message pop's up, the UI remains same.

**Expected behavior**

A snackbar with the message "email and password fields required to login" or error message with red icon in the email, password edit text.

**Screenshots**

ezgif com-video-to-gif-1

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | [Redmi note 5] |
|Android Version| [Oreo 8.1]  |

- [x] Yes

Cool-fire:
Fixes #927 

Changes: added TextInputlayout error with message when entered email is wrong

Screenshots for the change:

valid

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
@@ -13,9 +13,7 @@ import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.navigation.Navigation.findNavController
 import com.google.android.material.snackbar.Snackbar
-import kotlinx.android.synthetic.main.fragment_login.email
-import kotlinx.android.synthetic.main.fragment_login.loginButton
-import kotlinx.android.synthetic.main.fragment_login.password
+import kotlinx.android.synthetic.main.fragment_login.*
 import kotlinx.android.synthetic.main.fragment_login.view.email
 import kotlinx.android.synthetic.main.fragment_login.view.loginCoordinatorLayout
 import kotlinx.android.synthetic.main.fragment_login.view.forgotPassword
@@ -133,6 +131,7 @@ class LoginFragment : Fragment() {
     }
 
     private fun onEmailEntered(enable: Boolean) {
+        if (!enable) textInputLayoutEmail.error = getString(R.string.email_error) else textInputLayoutEmail.error = null
         rootView.loginButton.isEnabled = enable
         rootView.forgotPassword.isVisible = enable
     }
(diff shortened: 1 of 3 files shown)


Cool-fire:
@nikit19 updated the changes, now error shows only after user completed the writing.

please review the changes.

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -13,9 +14,7 @@ import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.navigation.Navigation.findNavController
 import com.google.android.material.snackbar.Snackbar
-import kotlinx.android.synthetic.main.fragment_login.email
-import kotlinx.android.synthetic.main.fragment_login.loginButton
-import kotlinx.android.synthetic.main.fragment_login.password
+import kotlinx.android.synthetic.main.fragment_login.*
```
kshitijsuri90: Wildcard imports are not allowed.

Cool-fire:
@kshitijsuri90 updated gif is at the top of conversation. Please check it there.

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
kshitijsuri90: Why marking as resolved without commit?

Cool-fire: @kshitijsuri90 is everything correct except that wildcard import.
Will change that wildcard import and will update the commit with changes.

kshitijsuri90: Cool. No need to resolve it now. It will automatically be resolved when you make the changed commit.

Cool-fire: Okay thanks @kshitijsuri90

kshitijsuri90:
IMO why not use the on focus change property of edit text to see of valid email is entered or not instead of using a timer?

Cool-fire:
@kshitijsuri90 IMO it is better to use a timer because it gives error after user stops writing not when user changes focus from the edittext. I think that's what @nikit19 needs.

Comments appreciated

Issue #1014, opened 2019-02-03
improving the layout of about_event fragment


pranavpandey1998official:
currently, it is not very readable in landscape mode the main content occupies less than 1/4  of screen

**Screenshots**
![device-2019-02-03-143256](https://user-images.githubusercontent.com/44601530/52174902-dad35d80-27c0-11e9-8372-36925d207313.png)

**Additional context**

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

anhanh11001:
Can I work on this?

pranavpandey1998official:
I am working on it

theLimitBreaker:
@nikit19 @iamareebjamal In landscape mode content is visible and properly scrollable . And fit to screen. Its irrelevant now

PR #1027, opened 2019-02-04
fix: improved the layout of about_event fragment

pranavpandey1998official:
introduced scroolView and removed old collapsingtoolbarlayout and added function to set action bar title

fix: #1014 

screen shot

about2
about1

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/about/AboutEventFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/about/AboutEventFragment.kt
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.general.about
 
+import android.os.Build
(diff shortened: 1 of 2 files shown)


iamareebjamal:
Please create an issue to strip all tags from the content coming from backend and then show it

pranavpandey1998official:
@iamareebjamal  mean html tags

iamareebjamal:
Yes it is but we can strip them idempotently, for new and old events since there is no timeline when or if server will solve this issue

iamareebjamal:
Both of above issues talk about displaying HTML and links. We are not going to do that. We will be changing HTML to plain text

pranavpandey1998official:
@iamareebjamal @nikit19 please review it

PR #1030, opened 2019-02-04
fix: multiple snackbars in attendee details

### LINKED ISSUE #1024: Snackbar pops up twice in Attendee Details
**Describe the bug**
When nothing is filled in attendee details, the snackbar pops up twice.
**To Reproduce**
Event - tickets - register
Don't fill anything and click register.

**Expected behavior**
Snackbar should appear once.

**Screenshots**
videotogif_2019 02 04_12 22 39

**Smartphone Info:**
Redmi note5 pro 8.1

- [x] Yes

shiv07tiwari:
Fixes #1024 

Changes:Added a count of how many attendee details are invalid, and send a snackbar message only for the first non valid attendee.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
@@ -68,6 +68,7 @@ class AttendeeViewModel(
     val year = ArrayList<String>()
     val attendees = ArrayList<Attendee>()
     val cardType = ArrayList<String>()
+    var isAllDetailsFilled = true
 
     private var createAttendeeIterations = 0
     var country: String? = null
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -162,11 +163,12 @@ class AttendeeViewModel(
 
     private fun createAttendee(attendee: Attendee, totalAttendee: Int) {
         if (attendee.email.isNullOrEmpty() || attendee.firstname.isNullOrEmpty() || attendee.lastname.isNullOrEmpty()) {
-            mutableMessage.value = "Please fill in all the fields"
+            detailsNotFilledCount++
+            if (detailsNotFilledCount ==1)
+                mutableMessage.value = "Please fill in all the fields"
```
iamareebjamal: Why?

shiv07tiwari: Added this so that the user gets Snackbar message only once, instead of for all the attendees. Right now the user gets Snackbar for all those attendees whose details aren't filled. @iamareebjamal

iamareebjamal: Then just don't set the message value in this method

shiv07tiwari: @iamareebjamal Then where should I add the message value ?
Can I run a loop in createAttendees method to check if all the details are filled, and then only call createAttendee method ?
@nikit19 I will fix the travis error.

iamareebjamal: Yes, add a check in AttendeeViewModel only

shiv07tiwari:
@iamareebjamal Made the implementation changes.

shiv07tiwari:
@nikit19 Made the changes with travis fix.

shiv07tiwari:
@nikit19 @liveHarshit Kindly review this PR.

PR #1041, opened 2019-02-05
fix: Use okhttp3 Interceptor in place of Authenticator for network module

### LINKED ISSUE #1006: URGENT: Check if first requests are sent without authentication token
I was reading the implementation of Authenticator of OkHTTP3 and it seems it only calls the callback to set authentication headers IF the server fails and give 401. Please check if this is true since this can cause unauthenticated information to be shown to the user for endpoints which support both authenticated and unauthenticated users

@liveHarshit @GOVINDDIXIT Can you please check this?

liveHarshit:
Fixes #1006

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/RequestAuthenticator.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/RequestAuthenticator.kt
@@ -1,18 +1,19 @@
 package org.fossasia.openevent.general.auth
 
-import okhttp3.Authenticator
-import okhttp3.Request
+import okhttp3.Interceptor
 import okhttp3.Response
-import okhttp3.Route
 
-class RequestAuthenticator(private val authHolder: AuthHolder) : Authenticator {
+class RequestAuthenticator(private val authHolder: AuthHolder) : Interceptor {
 
-    override fun authenticate(route: Route?, response: Response): Request? {
-        val authorization: String? = authHolder.getAuthorization()
-        return if (response.request().header("Authorization") == null && authorization != null) {
-            response.request().newBuilder()
-                    .header("Authorization", authorization)
-                    .build()
-        } else null // Give up, we already tried authenticating
+    override fun intercept(chain: Interceptor.Chain): Response? {
+        val authorization = authHolder.getAuthorization()
+        val original = chain.request()
+        return if (authorization != null) {
+            val request = original.newBuilder()
+                .header("Authorization", authorization)
+                .build()
+            chain.proceed(request)
+        } else
+            chain.proceed(original)
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/RequestAuthenticator.kt
```diff
@@ -1,18 +1,19 @@
...
+import okhttp3.Interceptor
 import okhttp3.Response
-import okhttp3.Route
 
-class RequestAuthenticator(private val authHolder: AuthHolder) : Authenticator {
+class RequestAuthenticator(private val authHolder: AuthHolder) : Interceptor {
 
-    override fun authenticate(route: Route?, response: Response): Request? {
+    override fun intercept(chain: Interceptor.Chain): Response? {
         val authorization: String? = authHolder.getAuthorization()
```
iamareebjamal:
```suggestion
        val authorization = authHolder.getAuthorization()
```

app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
```diff
@@ -153,7 +153,7 @@ val networkModule = module {
         objectMapper
     }
 
-    single { RequestAuthenticator(get()) as Authenticator }
+    single { RequestAuthenticator(get()) as Interceptor }
```
iamareebjamal: No need to cast IMO

app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
```diff
@@ -165,7 +165,6 @@ val networkModule = module {
                 .addInterceptor(
                         HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY)
                 )
-                .authenticator(get())
                 .build()
```
iamareebjamal: You have not added the interceptor here. What's the point then?

app/src/main/java/org/fossasia/openevent/general/auth/RequestAuthenticator.kt
```diff
@@ -1,18 +1,19 @@
...
-        return if (response.request().header("Authorization") == null && authorization != null) {
-            response.request().newBuilder()
-                    .header("Authorization", authorization)
-                    .build()
-        } else null // Give up, we already tried authenticating
+    override fun intercept(chain: Interceptor.Chain): Response? {
+        val authorization = authHolder.getAuthorization()
+        val original = chain.request()
+        val request = original.newBuilder()
+            .header("Authorization", authorization.toString())
```
iamareebjamal: Why toString?

liveHarshit: It is showing error while running `Authorization == null` and network module is not working. Because here `authorization` type is nullable String.

iamareebjamal: Only run this if (!Utils.isEmpty), problem solved

liveHarshit: Updated

PR #1046, opened 2019-02-06
feat: add landscape orientation resources for welcome screen

### LINKED ISSUE #1044: Add landscape orientation resources for welcome screen
Add landscape orientation resources for the welcome screen.

**Screenshots**
welcome_screen

**Additional context**
Since the app supports landscape orientation, there should be resources(layouts/strings etc) for the same. Some screens are good without needing separate resources but some are unusable without them. Welcome screen's current layout doesn't work in landscape orientation.

- [X] Yes
- [ ] No
- Other:

AnEnigmaticBug:
Fixes #1044 

Changes: This adds a landscape layout and a landscape strings.xml file. This makes the welcome screen usable in landscape configuration.

Screenshots for the change:
Before
welcome_screen
After
welcome_screen_new

### DIFF
--- a/app/src/main/res/layout-land/fragment_welcome.xml
+++ b/app/src/main/res/layout-land/fragment_welcome.xml
@@ -7,7 +7,7 @@
     tools:context="org.fossasia.openevent.general.WelcomeFragment">
 
     <TextView
-        android:id="@+id/textView"
+        android:id="@+id/heading"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginStart="@dimen/layout_margin_large"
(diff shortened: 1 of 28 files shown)


app/src/main/res/layout-land/fragment_welcome.xml
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="org.fossasia.openevent.general.WelcomeFragment">
+
+    <TextView
+        android:id="@+id/textView"
```
nikit19: Add meaningful variable names

nikit19: elsewhere too

AnEnigmaticBug: Ok. I'll add them

AnEnigmaticBug: @nikit19 the corresponding portrait layout doesn't have variable names for the same. Do I add them only in the landscape layout?

AnEnigmaticBug: If I add ids which aren't there in portrait mode layout, then someone can potentially try using them in Kotlin. Which may be error-prone. The id would exist only in landscape mode and not in portrait mode. If I should add ids, shouldn't I add them for both of them?

AnEnigmaticBug: @nikit19  I've made the changes

iamareebjamal:
What are the differences in landscape and portrait layout?

AnEnigmaticBug:
@iamareebjamal the landscape layout uses ConstraintLayout. This makes it have a flatter view hierarchy and allows the layout to work predictably on many devices. In the portrait layout, the globe image's size is fixed. But in the landscape layout, the image's size depends upon the space available. This prevents overlapping of UI elements.

AnEnigmaticBug:
@nikit19 do you require any further changes?

app/src/main/res/layout-land/fragment_welcome.xml
AnEnigmaticBug: @nikit19 can you please review my changes?

Issue #1048, opened 2019-02-06
Data retention on orientation change for all activities.


angmas1:
**Describe the bug**
Many activities loose data on orientation change.

**To Reproduce**
Rotate device in any activity after entering data.

**Expected behavior**
Data shouldn't be lost on device rotation.

**Would you like to work on the issue?**
Yes

shiv07tiwari:
@angmas1 A lot of people are already working on this issue talking up different activities. 
https://github.com/fossasia/open-event-android/issues/1016
https://github.com/fossasia/open-event-android/issues/967

Issue #1005, opened 2019-02-02
App crashes when no fields filled after screen rotation in Attendee details.


shiv07tiwari:
**Describe the bug**
Generally when no data is entered in Attendee details, it shows a snackbar to enter details.
But when screen is rotated, then instead of snackbar, the app crashes.

**To Reproduce**
Open any event -> Go to Tickets -> Register -> Enter details.
Dont fill anything, rotate the screen. Click on Register.

**Expected behavior**
It should show a snackbar to enter details.

**Screenshots**
![videotogif_2019 02 02_17 26 21](https://user-images.githubusercontent.com/38163725/52163968-3db8ec00-2710-11e9-9896-2ccbecc0261f.gif)

**Logs**

02-02 17:31:27.816 6646-6646/com.eventyay.attendee E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.eventyay.attendee, PID: 6646
    java.lang.ArrayIndexOutOfBoundsException: length=10; index=-1
        at java.util.ArrayList.get(ArrayList.java:439)
        at org.fossasia.openevent.general.attendees.AttendeeFragment$onCreateView$19.onClick(AttendeeFragment.kt:387)
        at android.view.View.performClick(View.java:6304)
        at android.view.View$PerformClick.run(View.java:24803)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6600)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:518)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:821)

**Smartphone Info:**
Redmi Note5 pro, Oreo 8.1

**Would you like to work on the issue?**
- [x] Yes
- [ ] No

pranavpandey1998official:
will be solved by #967

angmas1:
unable to reproduce

Issue #1004, opened 2019-02-02
Attendee details loses data on screen rotation.


shiv07tiwari:
**Describe the bug**
The first name and last name in Attendee details is lost when screen is rotated.
**To Reproduce**
Open any event -> Go to Tickets -> Register ->  Enter details.
Rotate the screen, name and last name is lost.

**Expected behavior**
Data should retain in screen rotation.

**Screenshots**
![videotogif_2019 02 02_17 16 17](https://user-images.githubusercontent.com/38163725/52163897-0138c080-270f-11e9-8a65-e8f0e5ecdacd.gif)

**Smartphone Info:**
Redmi Note 5 pro
 Oreo 8.1

**Would you like to work on the issue?**
- [x] Yes
- [ ] No

pranavpandey1998official:
@shiv07tiwari same as #967 and i am already working on the issue

shiv07tiwari:
@pranavpandey1998official Sorry I could not see that.
Are you also fixing the Ticket Details section ?
I would like to fix that.

anhanh11001:
I'm closing this as this is already fixed in #1810

Issue #967, opened 2019-01-27
attendees ticket info not retaining after screen rotation


pranavpandey1998official:
**Describe the bug**
attendees ticket info not retaining after screen rotation
and also no of ticket block doubles after rotation of screen 
**To Reproduce**
1. Go to attendees fragment by applying for any ticket and select 2 or more ticket in ticket detail
2. now fill the attendees ticket form then rotate the screen

**Screenshots**
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/44601530/51798491-2625c300-2239-11e9-8969-fa8f41641f79.gif)

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | [redmi not 4]|
|Android Version|[7.0] |

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

kshitijsuri90:
This is similar to issue #965. Will be solved with its PR. Will include this issue in the PR

pranavpandey1998official:
sorry i am working on it

PR #1049, opened 2019-02-06
fix: Data retention on orientation change for all activities.

### LINKED ISSUE #1016: User Details and Image is lost in screen rotation
**Describe the bug**
When the user edits his/her profile, then the entered data and loaded image is not retained when the screen is rotated.

**To Reproduce**
Profile -> Edit -> Enter details, select an image and rotate the screen.

**Expected behavior**
Entered data and image should retain

**Screenshots**
videotogif_2019 02 03_17 12 19

**Smartphone Info:**
Redmi Note5 pro 8.1 Oreo

- [x] Yes
- [ ] No

angmas1:
Fixes #1016 #1005 #967 #1004 #652 #1048 

Changes: Changed AndroidManifest.xml to have android:configChanges="orientation|screenSize" which prevents call onCreate when orientation is changed.

Screenshots for the change:
ezgif com-video-to-gif
ezgif com-video-to-gif 1

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -16,6 +16,7 @@
         android:theme="@style/AppTheme">
         <activity
             android:name=".MainActivity"
+            android:configChanges="orientation|screenSize"
             android:windowSoftInputMode="adjustPan"
             android:theme="@style/SplashTheme">
             <intent-filter>


angmas1:
> Wrong
Why? Please elaborate.
This solves all the problems.

iamareebjamal:
Try to change the landscape layout of the activity and see if it works

angmas1:
Specify one or more configuration changes that the activity will handle itself. If not specified, the activity will be restarted if any of these configuration changes happen in the system. Otherwise, the activity will remain running and its {@link android.app.Activity#onConfigurationChanged Activity.onConfigurationChanged} method called with the new configuration.

Description of this property in Android Studio.

According to this orientation change will be ignored only if it's not specified in the activity.

angmas1:
I checked the code but couldn't find any instance where specific layout was present for landscape mode. Please let me know if you find any. All fragments that I checked worked as they used to before.

shiv07tiwari:
@angmas1 A lot of people are already working on this issue talking up different activities.
#1016
#967

iamareebjamal:
I asked you to try to change the orientation layout of the fragment. The answer is you can't

The correct solution is to correctly use ViewModels and LiveData

PR #1058, opened 2019-02-08
fix: Disable text fields and buttons during progress.

### LINKED ISSUE #1045: Disable text fields and button during progress
**Describe the bug**
Disable text field and buttons during progress in login and sign up fragments. Because during sign in or smart authentication it is not a good user experience to make edit field accessible.

Better solution is to use an overlapping progress bar. 

- Other: Open for all

angmas1:
Fixes #1045 

Changes: When sign-up/login buttons are clicked, the button gets replaced with progress bar and the edit text fields get disabled.

Screenshots for the change:
ezgif com-video-to-gif 1
ezgif com-video-to-gif


app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -14,9 +14,7 @@ import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.navigation.Navigation.findNavController
 import com.google.android.material.snackbar.Snackbar
-import kotlinx.android.synthetic.main.fragment_login.email
-import kotlinx.android.synthetic.main.fragment_login.loginButton
-import kotlinx.android.synthetic.main.fragment_login.password
+import kotlinx.android.synthetic.main.fragment_login.*
```
GOVINDDIXIT: avoid wild card imports

angmas1: okay

angmas1:
@nikit19 corrected please review.

liveHarshit:
Like that -  no need to handle visibality and disable text fields. 
<p align="center">
 
</p>

angmas1:
Okay

angmas1:
> Like that - no need to handle visibality and disable text fields.
> 
> 

image

Done!

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -56,6 +56,11 @@ class LoginFragment : Fragment() {
         setHasOptionsMenu(true)
         showSnackbar()
 
+        val dialog = Dialog(context)
+        dialog.requestWindowFeature(Window.FEATURE_NO_TITLE)
+        dialog.setCancelable(false)
+        dialog.setContentView(R.layout.dialog_progress)
```
liveHarshit: Can you move dialog logic to utils, so we can display it anytime by passing context?

iamareebjamal:
The dialog is too large

angmas1:
Okay, moving dialog to utils and making it smaller.

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
angmas1: done

angmas1:
image

image

angmas1:
ezgif com-video-to-gif 1

ezgif com-video-to-gif

app/src/main/java/org/fossasia/openevent/general/utils/ProgressDialog.kt
```diff
@@ -0,0 +1,20 @@
...
+        dialog.setCancelable(false)
+        dialog.requestWindowFeature(Window.FEATURE_NO_TITLE)
+        dialog.setContentView(R.layout.dialog_progress)
+    }
+
+    fun showOrDismiss(show: Boolean) {
+        if (show && !dialog.isShowing) dialog.show()
+        else if (!show && dialog.isShowing) dialog.dismiss()
+    }
+}
```
liveHarshit: Utils mean in utils.kt file.

liveHarshit: Create a method for the same in utils, no need to create a new file.

angmas1: That won't work because loading dialog is set as non-cancel able, so I have to cancel each dialog box too. So, if I make it into a function then I won't be able to cancel the dialog created on previous call. Hence, I need to create an object of the progressdialog for the entire fragment. This solves the issue. If I make the dialog box cancel able, then the user can just click outside the dialog box which will make it pointless as it won't stop user intervention on login/signup

liveHarshit: In utils - 
```java
fun showProgressDialog(context: Context, show: Boolean) {
    private val dialog = Dialog(context)
    dialog.setCancelable(false)
    dialog.requestWindowFeature(Window.FEATURE_NO_TITLE)
    dialog.setContentView(R.layout.dialog_progress)
    if (show && !dialog.isShowing) dialog.show() else dialog.dismiss()
}
```

And just call the method - 
`Utils.showProgressDialog(requireContext(), true)`
`Utils.showProgressDialog(requireContext(), false)`

Hope it should work.

angmas1: No, it doesn't. That's what I wrote in beginning. The thing is, when this is called it creates a val dialog. So, dialog.isShowing will always return false, because you are referring to the recently created dialog. Hence, you will never be able to close a dialog which was created on previous call and keep on creating new dialogs.

liveHarshit: got it!

liveHarshit: Then you can use it like that - 

In fragment - 
`val dialog = Utils.progressDialog(requireContext())`

In Utils.kt - 

```java
fun progressDialog(context: Context) : Dialog {
    val dialog = Dialog(context)
    dialog.setCancelable(false)
    dialog.requestWindowFeature(Window.FEATURE_NO_TITLE)
    dialog.setContentView(R.layout.dialog_progress)
   return dialog
}
```

And visibility can handle inside the fragment.

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -81,13 +81,13 @@ class LoginFragment : Fragment() {
         loginViewModel.progress
             .nonNull()
             .observe(this, Observer {
-                handleProgressBar(it)
+            progressDialog.showOrDismiss(it)
```
liveHarshit: Wrong indentation.

app/src/main/java/org/fossasia/openevent/general/utils/ProgressDialog.kt
angmas1: Can I just move the class to utils.kt?

angmas1: Because making a class will be easier when reusing code.

angmas1: Moreover, class will use less memory.

liveHarshit: IMO, creating a new class only for a progress dialog is not a good idea, when you can just use a common method instead to remove duplication.

liveHarshit: @iamareebjamal @nikit19 Please make me correct.

iamareebjamal: > Moreover, class will use less memory

1 class will use less memory than no class?

angmas1: What @liveHarshit is suggesting is making a function which initializes the dialog(just like a class constructor.). And handle the logic in the place where the function is called. This is why I'm suggesting a class so that we can handle the logic and initialization in one place. It will be easier then for someone else to use it any where else. Moreover, class will use less memory as in when an object of progressDialog is made, it will use space for one object. Whereas when that initialize method is called it creates an object and then passes it to the object of calling function. Which means for a small amount of time, two objects are in memory. One of calling function and one of initialize function itself. Please correct me if I am wrong.

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
angmas1: Thanks for pointing out. Will correct.

app/src/main/java/org/fossasia/openevent/general/utils/ProgressDialog.kt
iamareebjamal: If two classes are invoking two dialogs, the amount of classes are going to be in memory in any case. There is no need to create a class. Make a function which returns a dialog

angmas1: Okay

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
angmas1: Corrected

app/src/main/java/org/fossasia/openevent/general/utils/ProgressDialog.kt
angmas1: Done

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -81,14 +81,14 @@ class LoginFragment : Fragment() {
         loginViewModel.progress
             .nonNull()
             .observe(this, Observer {
-                handleProgressBar(it)
+                handleProgressBar(it, progressDialog)
```
liveHarshit: Please declare progressDialog as `private lateinit var` before onCreateView and initialize it in onCreateView. So that there is no need to pass progressDialog.

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -171,6 +171,11 @@ class LoginFragment : Fragment() {
         smartAuthViewModel.requestCredentials(activity)
     }
 
+    private fun handleProgressBar(show: Boolean, dialog: Dialog) {
```
iamareebjamal: No need of this function

iamareebjamal: Create an extension function on Dialog `show(boolean) which does this

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
liveHarshit: As @iamareebjamal suggested, no need to create function

app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
angmas1: Okay

angmas1: Done

app/src/main/java/org/fossasia/openevent/general/auth/SignUpFragment.kt
```diff
@@ -76,8 +74,8 @@ class SignUpFragment : Fragment() {
         signUpViewModel.progress
             .nonNull()
             .observe(this, Observer {
-                rootView.progressBarSignUp.isVisible = it
-                signUpButton.isEnabled = !it
+                if (it && !progressDialog.isShowing) progressDialog.show()
+                else if (!it && progressDialog.isShowing) progressDialog.dismiss()
```
iamareebjamal: Use common method

angmas1: done

app/src/main/java/org/fossasia/openevent/general/auth/SignUpFragment.kt
```diff
@@ -130,6 +128,11 @@ class SignUpFragment : Fragment() {
         Snackbar.make(rootView, R.string.logged_in_automatically, Snackbar.LENGTH_SHORT).show()
     }
 
+    private fun Dialog.show(show: Boolean) {
+        if (show && !this.isShowing) this.show()
+        else if (!show && this.isShowing) this.dismiss()
+    }
```
iamareebjamal: Why are there 2 methods in 2 activities doing the same thing?

app/src/main/res/layout/dialog_progress.xml
```diff
@@ -0,0 +1,21 @@
...
+
+        <TextView
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:paddingRight="@dimen/padding_extreme_large"
+            android:text="Loading..."
+            android:textSize="@dimen/text_size_large" />
+
+</LinearLayout>
```
iamareebjamal: Why do we even need this. Isn't progress dialog good enough?

angmas1: To setcontentview for dialog.

iamareebjamal: Why do we need that?

angmas1: This is providing layout for the dialog loading dialog that pops up.

iamareebjamal: Again, why do we need a custom layout?

angmas1: I don't know any other way of adding a progressBar to a dialog body other than making a custom layout.

app/src/main/java/org/fossasia/openevent/general/auth/SignUpFragment.kt
angmas1: Resolved

app/src/main/res/layout/dialog_progress.xml
iamareebjamal: https://developer.android.com/reference/android/app/ProgressDialog

angmas1: Okay, will implement this.

angmas1: done

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
```diff
@@ -44,6 +46,21 @@ object Utils {
             .show()
     }
 
+    fun progressDialog(context: Context?): ProgressDialog {
+        /**Initializes progress dialog.*/
+        val dialog = ProgressDialog(context)
+        dialog.setCancelable(false)
+        dialog.requestWindowFeature(Window.FEATURE_NO_TITLE)
```
iamareebjamal: No need

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
```diff
@@ -44,6 +46,21 @@ object Utils {
...
+        dialog.setMessage("Loading...")
+        return dialog
+    }
+
+    fun ProgressDialog.show(show: Boolean) {
+        /**Extends dialog.show()*/
+        if (show && !this.isShowing) this.show()
```
iamareebjamal: Remove  && !this.isShowing

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
```diff
@@ -44,6 +46,21 @@ object Utils {
...
+        else if (!show && this.isShowing) this.dismiss()
```
iamareebjamal: Change to else

app/src/main/res/values/dimens.xml
```diff
@@ -26,6 +26,7 @@
     <dimen name="padding_medium">8dp</dimen>
     <dimen name="padding_large">16dp</dimen>
     <dimen name="padding_extra_large">32dp</dimen>
+    <dimen name="padding_extreme_large">128dp</dimen>
```
iamareebjamal: No need

app/src/main/res/values/dimens.xml
```diff
@@ -49,6 +50,7 @@
     <dimen name="app_bar_height">250dp</dimen>
     <dimen name="fab_margin">16dp</dimen>
     <dimen name="text_margin">16dp</dimen>
+    <dimen name="progress_bar_size">64dp</dimen>
```
iamareebjamal: No need

angmas1: done.

app/src/main/res/values/dimens.xml
angmas1: done.

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
angmas1: done.

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
angmas1: done.

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
angmas1: done.

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
```diff
@@ -44,6 +45,20 @@ object Utils {
             .show()
     }
 
+    fun progressDialog(context: Context?): ProgressDialog {
+        /**Initializes progress dialog.*/
```
iamareebjamal:
```suggestion
```

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
```diff
@@ -44,6 +45,20 @@ object Utils {
...
+        val dialog = ProgressDialog(context)
+        dialog.setCancelable(false)
+        dialog.setMessage("Loading...")
+        return dialog
+    }
+
+    fun ProgressDialog.show(show: Boolean) {
+        /**Extends dialog.show()*/
```
iamareebjamal:
```suggestion
```

PR #1062, opened 2019-02-08
fix: Social Links and Similar Events get refetched when event is favorited

### LINKED ISSUE #256: Social Links and Similar Events get refetched when event is favorited
Social Links and Similar events are being refetched when event is favorited using toolbar icon or fab icon present in Similar events. This should not occur

Working on it :+1:

angmas1:
Fixes #256 

Changes: Stopped re fetching of similar events and organization details on setting favorite.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -70,7 +70,7 @@ class EventDetailsFragment : Fragment() {
     private val LINE_COUNT: Int = 3
     private var menuActionBar: Menu? = null
     private var title: String = ""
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -70,6 +70,7 @@ class EventDetailsFragment : Fragment() {
     private val LINE_COUNT: Int = 3
     private var menuActionBar: Menu? = null
     private var title: String = ""
+    private var runonce: Boolean = true
```
dreadpool2: use camel case (runOnce)

angmas1: Okay

PR #1067, opened 2019-02-09
fix: multiple popups in login screen on device rotation

### LINKED ISSUE #1066: bug: rotating device in login screen leads to multiple login popups
**Describe the bug**
When the login screen starts, a popup asking the user to choose an email is shown. The issue is that when the device is rotated, it is shown again. This leads to multiple popups placed on top of each other.

**To Reproduce**
1. Go to login screen. The popup should appear.
2. Without dismissing/closing the popup, rotate the device.
3. Repeat this 1-2 more times.
4. You should be seeing 2-3 popups in front of you(try dismissing them).

**Expected behavior**
Ideally, rotating the screen shouldn't prompt the user again for showing the popup. Multiple popups shouldn't be shown in any event.

**Screenshots**
Since popups appear on top of each other, they won't be visible separately in a screenshot.

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Redmi Note 4|
|Android Version|7.0|

- [X] Yes
- [ ] No

AnEnigmaticBug:
Fixes #1066 

Changes: Earlier, rotating the device would cause multiple popups to be shown if the user didn't dismiss the earlier ones. This was because the popup was shown whenever onStart() of the LoginFragment was called. But onStart() is called everytime on device rotation, leading to this bug.

Now, a boolean tracks whether the popup has been shown before.

Screenshots for the change:
There is no visual difference. The popups used to overlap each other completely, making it impossible to distinguish between them via screenshots.

### DIFF
--- a/app/src/playStore/java/org/fossasia/openevent/general/search/SmartAuthViewModel.kt
+++ b/app/src/playStore/java/org/fossasia/openevent/general/search/SmartAuthViewModel.kt
@@ -22,6 +22,7 @@ const val RC_CREDENTIALS_READ = 2
 const val RC_CREDENTIALS_SAVE = 3
 
 class SmartAuthViewModel : ViewModel() {
+    private var requestedCredentialsEarlier = false
     private lateinit var credentialsClient: CredentialsClient
     private lateinit var signInClient: GoogleSignInClient
     private lateinit var thisActivity: Activity
@@ -45,13 +46,15 @@ class SmartAuthViewModel : ViewModel() {
     }
 
     fun requestCredentials(activity: Activity?) {
-        if (activity == null) return
+        if (activity == null || requestedCredentialsEarlier) return
         val crBuilder = CredentialRequest.Builder()
             .setPasswordLoginSupported(true)
         crBuilder.setAccountTypes(IdentityProviders.GOOGLE)
         mutableProgress.value = true
         credentialsClient.request(crBuilder.build()).addOnCompleteListener(
             OnCompleteListener<CredentialRequestResponse> { task ->
+                requestedCredentialsEarlier = true
+
                 mutableProgress.value = false
                 if (task.isSuccessful) {
                     mutableId.value = task.result?.credential?.id


app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
```diff
@@ -152,7 +152,9 @@ class LoginFragment : Fragment() {
 
     override fun onStart() {
         super.onStart()
-        smartAuthViewModel.requestCredentials(activity)
+        if (!smartAuthViewModel.requestedCredentialsEarlier) {
+            smartAuthViewModel.requestCredentials(activity)
+        }
```
iamareebjamal: Why are you exposing and accessing the variable from activity. Make the `requestedCredentialsEarlier` private

AnEnigmaticBug: The current method name of the request credentials method is "requestCredentials". I thought that the name implies that calling it will always request credentials by showing a popup. So, to make the fact that we only request credentials once more explicit, I exposed the variable public and used it in the Activity.

Can you please tell me what should I do?

iamareebjamal: Return from the viewmodel if it is requested more than once, why give control to the fragment?

AnEnigmaticBug: I'll push the changes right away

app/src/fdroid/java/org/fossasia/openevent/general/search/SmartAuthViewModel.kt
```diff
@@ -10,6 +10,7 @@ const val RC_CREDENTIALS_READ = 2
 const val RC_CREDENTIALS_SAVE = 3
 
 class SmartAuthViewModel : ViewModel() {
+
```
iamareebjamal: Unneeded

AnEnigmaticBug: I removed it

iamareebjamal:
Thank you for the detailed commit description

AnEnigmaticBug:
You're welcome :)

PR #1076, opened 2019-02-11
feat: make bottom nav icons more consistent

### LINKED ISSUE #1075: feat: make bottom nav icons more consistent
Right now, the icons in the bottom bar are a mix of filled and outlined icons. While it looks good enough right now, it can be made consistent and even better.

All bottom nav icons should be outlined.

**Screenshots**
Current icons:
bottom_nav_icons_current

auth_current

**Additional context**
Consistency in design elements can subtly improve the user experience.

- [X] Yes
- [ ] No

AnEnigmaticBug:
Fixes #1075

Changes: Earlier, the bottom nav icons consisted of both filled and outline style
icons. This led to a slightly inconsistent look.

Now, the icons are all outline icons. Also, the icon for tickets has
been changed to a (IMHO) slightly more appropriate one.

Screenshots for the change:
current:
main_current
auth_current

updated:
main_updated
auth_updated


AnEnigmaticBug:
@liveHarshit  Ok. I'll look up an icon which is similar to the app icon and use it in vertical orientation. I just used this icon from material.io because other icons are also from material.io.

liveHarshit:
> @liveHarshit Ok. I'll look up an icon which is similar to the app icon and use it in vertical orientation. I just used this icon from material.io because other icons are also from material.io.

You can use other vector icon sites.

iamareebjamal:
It's fine

Issue #1082, opened 2019-02-12
chore: make different Koin modules for different flavors(fdroid and playstore)


AnEnigmaticBug:
**Describe the chore**
Right now, we're using just a single set of Koin modules.

While it works, we've to do some slightly hacky things. Some examples:
1. Even though the fdroid flavor doesn't have any location services, the `GeoLocationViewModel` class is there in the code.
2. `GeoLocationViewModel` is used in the `WelcomeFragment` in fdroid flavor. But the methods don't do anything. It doesn't make sense and looks confusing to a newcomer.
2. Both the flavors of `GeoLocationViewModel` have to be passed `LocationServices` and `Preference`. But those dependencies only make sense in the playstore flavor.

Keeping just a single set of modules means that not only we can't do anything about the code in welcome screen, but we'll have to write more such code in future.

If we put all the common modules together and just make different versions for places where the flavors differ appreciably, we can make the code more intuitive and clearer. As an example: maybe, we can rectify the above situation by not having a `GeoLocationViewModel` in fdroid at all.

**Would you like to work on the issue?**
- [X] Yes
- [ ] No

iamareebjamal:
Then we'll have to split some other fragment/component

AnEnigmaticBug:
I fully agree that we should avoid splitting whenever possible. We should only do it for a few screens to minimize any duplication. I just mean that if we set up different sets of modules, it gives us a greater flexibility in choosing which components to split.

Currently only welcome and search location screen differ between the two flavors.

Taking the welcome screen as an example, if we split just the `WelcomeFragment` then:
1. We can use the already existing `WelcomeViewModel` in the playstore flavor.
2. The fdroid flavor won't need a `ViewModel` because it doesn't display any data. The only interaction in fdroid flavor can be clicking the pick city button which is already done in the `Fragment` itself. So, it doesn't need a `ViewModel`.
3. Since we won't be needing a `ViewModel` for the fdroid welcome screen, we won't need to make an implementation of `LocationService` for fdroid. This will further reduce the code.

So, splitting `WelcomeFragment` will allow us to remove the current splitting of `WelcomeViewModel` and `LocationService`. This will result in a net gain since we've gone from 2 splittings to 1. The only thing hampering this is the usage of a single set of modules for both flavors.

This process may lead to an increase in overall code but will make the code clearer and less error-prone.

iamareebjamal:
How much increase will be there in code?

[...]

PR #1083, opened 2019-02-12
refactor: two ViewModels in the welcome screen into one ViewModel

### LINKED ISSUE #1073: feat: refactor the two ViewModels in the welcome screen into one ViewModel
Currently, there are two `ViewModel`s being used in the welcome screen(the `GeoLocationViewModel` and the `SearchLocationViewModel`).

In MVVM, each View is associated with only one `ViewModel`. The `ViewModel`'s task is to simply expose the properties which the View can bind to and persist during configuration changes. If there are two distinct functionalities(like persisting location name and doing geo-location work), there should be two classes(one for each) which are injected into this one `ViewModel`.

Doing this leads to the benefit that testing that the `ViewModel` alone ensures that the screen will work correctly if the View layer is correct. It also leads to a better separation of concerns since the actual functionality is encapsulated into separate classes and the `ViewModel` serves it's intended purpose which is to simply expose properties.

If another screen wants to use the geo-location feature, we can simply inject the class responsible for geo-location into it's `ViewModel`.

- [X] Yes
- [ ] No

AnEnigmaticBug:
Fixes #1073 

Changes: Earlier, there were two `ViewModel`s being used in `WelcomeFragment`.

In MVVM, a `View` is associated with only one `ViewModel`. To avoid duplication of logic, all distinct logic is split into smaller classes which can then be reused in different `ViewModel`s.

So now, `WelcomeFragment` has only one `ViewModel`(with two flavors: fdroid and playstore). Also, made a separate package for the welcome screen since its also a screen like others.

Screenshots for the change:
This is purely a refactoring. So, UI hasn't been changed in any way.

### DIFF
--- a/app/src/fdroid/java/org/fossasia/openevent/general/welcome/WelcomeViewModel.kt
+++ b/app/src/fdroid/java/org/fossasia/openevent/general/welcome/WelcomeViewModel.kt
@@ -1,0 +1,28 @@
+package org.fossasia.openevent.general.welcome
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import org.fossasia.openevent.general.common.SingleLiveEvent
+import org.fossasia.openevent.general.data.Preference
+import org.fossasia.openevent.general.search.LocationService
+
+class WelcomeViewModel(locationService: LocationService, preference: Preference) : ViewModel() {
+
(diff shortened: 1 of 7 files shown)


iamareebjamal:
Good work but now we have split two components. 1- Location component and 2- Welcome ViewModel

I'm going to merge but this is considered boilerplate

AnEnigmaticBug:
@iamareebjamal I'm trying to rectify this exact issue. Can you please read my new comment on #1082?

PR #1105, opened 2019-02-15
fix: show error message on empty email password

### LINKED ISSUE #927: No error message when user clicks on login button with email and password fields empty.

shiv07tiwari:
Fixes #927 

Changes: Removed the dependency of login button on email verification. IMO the login button should always be enabled so as to show the exact error to the user, otherwise functions like
private fun hasErrors(email: String?, password: String?): Boolean {
        if (email.isNullOrEmpty() || password.isNullOrEmpty()) {
            mutableError.value = "Email or Password cannot be empty!"
            return true
        }
        return false
    }
are of no use.
videotogif_2019 02 15_18 40 26

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/LoginFragment.kt
@@ -182,7 +182,6 @@ class LoginFragment : Fragment() {
     }
 
     private fun onEmailEntered(enable: Boolean) {
-        rootView.loginButton.isEnabled = enable
         rootView.forgotPassword.isVisible = enable
     }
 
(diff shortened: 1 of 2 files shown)


shiv07tiwari:
@liveHarshit @nikit19 @iamareebjamal Kindly review this PR

PR #1107, opened 2019-02-15
fix: Attach attendee fragment snackbars correctly. 

### LINKED ISSUE #995: Misplaced Snackbar in Attendee Details
**Describe the bug**

Snackbar shown in Attendee details are misplaced.

**To Reproduce**

Register for tickets

**Expected behavior**

They should be properly placed

**Logs**

**Screenshots**

snack_bar

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |One Plus 5|
|Android Version|  Pie 9.0.3  |

**Additional context**

- [x] Yes
- [ ] No
- Other:

angmas1:
Fixes #995 

Changes: Placed snackbar correctly. Also removed redundant and not needed snackbars. That is, previously 3 snackbars were created:
1. Attendees created successfully.
2. Order created successfully

Replaced these with only one snackbar. "Order created successfully". Because, I think user won't be interested in knowing attendees created successfully. Didn't remove the error snackbar for the same though. So that correct error is shown.

Screenshots for the change:
ezgif com-video-to-gif


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -183,7 +183,6 @@ class AttendeeViewModel(
                 if (attendees.size == totalAttendee) {
                     loadTicketsAndCreateOrder()
                     mutableIsAttendeeCreated.value = true
-                    mutableMessage.value = "Attendees created successfully!"
```
iamareebjamal: Why?

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -262,7 +261,6 @@ class AttendeeViewModel(
                     mutableProgress.value = false
                 }.subscribe({
                     orderIdentifier = it.identifier.toString()
-                    mutableMessage.value = "Order created successfully!"
```
iamareebjamal: Why?

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
angmas1: Why would the user want to know that attendees were created successfully. Multiple snackbars don't look good. Instead f showing snackbar at this stage, I show snackbar when order created. But haven't removed error message snackbars, so if attendees are not created successfully, user snackbar saying same will show up.

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
angmas1: This shows up twice because it's the function it calls also has same sanckbar.

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
iamareebjamal: `Why would the user want to know that attendees were created successfully`

Why wouldn't they?

Change it so that it is only shown once

angmas1: okay

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
angmas1: This is okay right?

angmas1: Changed it slightly because, in case of free tickets, confirmorderstatus is called which gives order created successfully snackbar. So changed it so that snackbar is shown when paid ticket because paid ticket has no such snackbar. Removes repetition of snackbar "Order created successfully!" in case of free tickets.

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
angmas1: done.

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
iamareebjamal: We want to show snackbar in both free and paid cases

angmas1: It is shown in both cases. In free it was showing twice.

angmas1: In case of free, confirmorderstatus is being called which shows same snackbar

PR #1148, opened 2019-02-22
fix: "See more" button disappear #1337

### LINKED ISSUE #1137: "See more" disappears when click back button from bottom navigation bar
**Describe the bug**

In the Event Details, when clicking on "See more" to check out the About information and then click back button in the bottom navigation bar, the "See more" word disappear

**Expected behavior**
"See more" button stays the same

**Screenshots**

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | Nokia 7 Plus |
|Android Version| Android 9, API 28 |

<b>Yes</b>

anhanh11001:
Fixes #1137 

### Changes: 
When the EventDetailFragment comes back from AboutFragment, the text doesn't count the line because it has to render the text view first. So that's why "See more" button disappear, I changed it by using post method on the text view. It will start to check for lines of text to see whether to display "See more" or not only after rendering the text view

### Notes
I made this change in the #1140 pull request but I screwed in branching and using Git so I close that pull request and make new one here. Sorry for the inconvienience, I'm still new and learning to make pull request properly

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -176,11 +176,13 @@ class EventDetailsFragment : Fragment() {
         if (!event.description.isNullOrEmpty()) {
             setTextField(rootView.eventDescription, event.description)
 
-            if (rootView.eventDescription.lineCount > LINE_COUNT) {
-                rootView.seeMore.visibility = View.VISIBLE
-                // start about fragment
-                rootView.eventDescription.setOnClickListener(aboutEventOnClickListener)
-                rootView.seeMore.setOnClickListener(aboutEventOnClickListener)
+            rootView.eventDescription.post {
+                if (rootView.eventDescription.lineCount > LINE_COUNT) {
+                    rootView.seeMore.visibility = View.VISIBLE
+                    // start about fragment
+                    rootView.eventDescription.setOnClickListener(aboutEventOnClickListener)
+                    rootView.seeMore.setOnClickListener(aboutEventOnClickListener)
+                }
             }
         } else {
             aboutEventContainer.visibility = View.GONE


anhanh11001:
@liveHarshit please review when you have the time

anhanh11001:
@liveHarshit I understand, so I'll try to take it slow and be not craving for making PR fast. 
So for PR that's not reviewed but get out-of-date, should I update branch myself or let the reviewer do it? And with this PR, should any changes be made?

liveHarshit:
> should I update branch myself or let the reviewer do it

Both cases are fine

liveHarshit:
> should any changes be made?

LGTM

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -177,11 +177,13 @@ class EventDetailsFragment : Fragment() {
         if (!event.description.isNullOrEmpty()) {
             setTextField(rootView.eventDescription, event.description?.stripHtml())
 
-            if (rootView.eventDescription.lineCount > LINE_COUNT) {
-                rootView.seeMore.visibility = View.VISIBLE
-                // start about fragment
-                rootView.eventDescription.setOnClickListener(aboutEventOnClickListener)
-                rootView.seeMore.setOnClickListener(aboutEventOnClickListener)
+            rootView.eventDescription.post {
```
iamareebjamal: Why do we need post?

anhanh11001: so the reason why "see more" disappear is because rootView.eventDescription.lineCount is equal 0. From what I understand and research on StackOverflow, the method lineCount() is called only the view is rendered. I use post() method to check for lineCount after the view is rendered. Correct me if I am wrong in this PR. Thanks

PR #1152, opened 2019-02-22
fix: After coming back from result detail, search result come back to the top, not current position #1150

### LINKED ISSUE #1150: After coming back from result detail, search result come back to the top, not current position
**Describe the bug**
When users try click on a search result and then click back button, the search result doesn't stay at the same position but get to the top

** Images & Gif **
Gif link: https://i.ibb.co/Qmb3Cwv/video-1550686001.gif (file too large, cannot upload to issue)

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | Nokia 7 Plus|
|Android Version| Android 9 API 28 |

<b>Yes </b>

anhanh11001:
Fixes #1150 

Changes:
I checked if the view is already initialized. If it is, I reuse the view of search result, not reload the whole thing again. I also realized that the events fragment have the same problem so I make the same change

Screenshots for the change:

### DIFF
--- a/app/src/main/res/layout/fragment_events.xml
+++ b/app/src/main/res/layout/fragment_events.xml
@@ -103,7 +103,6 @@
                             android:scrollbars="vertical" />
                     </FrameLayout>
                 </LinearLayout>
-
                 <include layout="@layout/content_no_internet" />
             </LinearLayout>
         </androidx.coordinatorlayout.widget.CoordinatorLayout>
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/fragment_events.xml
```diff
@@ -103,7 +103,6 @@
                             android:scrollbars="vertical" />
                     </FrameLayout>
                 </LinearLayout>
-
```
liveHarshit: Revert it.

anhanh11001: I did

PR #1180, opened 2019-02-24
fix: Added Retry button

somenath1435:
Fixes #1175 

Changes: Added button instead of text(user can click on button)
AndroidX button is added.

Screenshots for the change:
Before:-
whatsapp image 2019-02-24 at 4 22 19 am

After:-
whatsapp image 2019-02-24 at 9 19 48 pm

### DIFF
--- a/app/src/main/res/layout/content_no_internet.xml
+++ b/app/src/main/res/layout/content_no_internet.xml
@@ -40,14 +40,15 @@
                 android:textSize="@dimen/text_size_medium"
                 tools:text="No Internet" />
 
-            <TextView
+            <androidx.appcompat.widget.AppCompatButton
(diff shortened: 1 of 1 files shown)


somenath1435:
@iamareebjamal I believe this button looks better than flat button.
Also, button shows that user can click on it, normal text may not feel like to be clicked by the user.

somenath1435:
@nitik19
Actually I added the button so that user can click on it.
User may not tap on plain text.

somenath1435:
@nikit19 I added the button so that user can click on it.
User may not click on plain text

somenath1435:
@iamareebjamal Shall I make some additional changes or shall I close this issue?

iamareebjamal:
Close

PR #1182, opened 2019-02-25
fix: After coming back from result detail, search result come back to the top, not current position #1150

anhanh11001:
Details: 
I checked if the view is already initialized. If it is, I reuse the view of the search result, not reload the whole thing again. I also realized that the events fragment have the same problem so I make the same change.

Besides, I prevent the change in action bar by setting the action bar first before rendering the rootView

Fixes #1150 

Screenshot:

PR #1183, opened 2019-02-25
fix: After coming back from result detail, search result come back to the top, not current position #1150

### LINKED ISSUE #1150: After coming back from result detail, search result come back to the top, not current position

anhanh11001:
Details: 
After coming from EventDetailsFragment, the SearchResultFragment doesn't retain the old position. Adding id for NestedScrollView will make Android automatically retain the current position

Fixes: #1150

### DIFF
--- a/app/src/main/res/layout/fragment_search_results.xml
+++ b/app/src/main/res/layout/fragment_search_results.xml
@@ -29,6 +29,7 @@
     </LinearLayout>
 
     <androidx.core.widget.NestedScrollView
+        android:id="@+id/searchResultsNestedScrollView"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:orientation="vertical"


liveHarshit:
Any special reason for closing PR #1182 and create a new one?

app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -88,8 +90,20 @@ class SearchLocationFragment : Fragment() {
...
+            val bundle = Bundle()
+            bundle.putBoolean(NEW_SEARCH, true)
+            val navOptions = NavOptions.Builder()
+                .setEnterAnim(R.anim.slide_in_left)
+                .setExitAnim(R.anim.slide_out_right)
+                .setPopEnterAnim(R.anim.slide_in_left)
+                .setPopExitAnim(R.anim.slide_out_right)
+                .setPopUpTo(R.id.eventsFragment, true)
+                .build()
+            Navigation.findNavController(rootView).navigate(R.id.eventsFragment, bundle, navOptions)
```
liveHarshit: Not a good solution. It was better earlier. Can you please find another way for it? Reset looks good.

anhanh11001: @liveHarshit I have updated a better solution. haha, I was stupid before, this is actually a much simpler issue.

app/src/main/res/layout/fragment_search_results.xml
```diff
@@ -29,6 +29,7 @@
     </LinearLayout>
 
     <androidx.core.widget.NestedScrollView
+        android:id="@+id/searchResultsNestedScrollView"
```
liveHarshit: Can you please explain, how defining an id solve the issue?

anhanh11001: @liveHarshit Well, in short, this is Android View State Storing Mechanism (to retain an amount of information/state if we provide it an id). So for instance, if you're having an EditText and some text in it, I believe the text is also retained after screen rotation if we provide it with an id.

So this is also the way current EventsFragment restoring its position after coming back from EventDetailsFragment.

PR #1189, opened 2019-02-26
fix: data loss on rotating EditProfileFragment

### LINKED ISSUE #1188: Data Retention on EditProfile when rotating device
**Describe the bug**
In EditProfileFragment, changes on EditText is not retained when rotating

**Screenshots**
Before rotating:

After rotating:

Yes

anhanh11001:
Details:
In here, I deleted the use of ProfileViewModel because it is only used to fetch the data in the beginning and set the edit text to original firstname/lastname. So whenever a user rotates the screen, a ViewModel is created again and fetch the data again -> editText follows those changes and keep resetting the text to the original firstname/lastname.

Instead, I initially pass the data from ProfileFragment so that we don't have to fetch the data from the server again. Text in the EditText is automatically retained now.

Fixes #1188 

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
@@ -38,7 +38,6 @@ import java.io.FileNotFoundException
 
 class EditProfileFragment : Fragment() {
 
-    private val profileViewModel by viewModel<ProfileViewModel>()
     private val editProfileViewModel by viewModel<EditProfileViewModel>()
     private lateinit var rootView: View
     private var permissionGranted = false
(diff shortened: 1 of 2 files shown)


anhanh11001:
I also believe the misuse of ViewModel in the view containing EditText like here or in Attendee Detail Fragment is the cause of data retention, which is described in some previous issues like #1004 . Correct me if I am wrong because I'm still learning about MVVC, so there might something I don't know or understand clearly about the ViewModel

iamareebjamal:
Don't delete the use of view model and check how to retain the data using view model

anhanh11001:
Here is the initial use of profileViewModel in this fragment
```
profileViewModel.user
            .nonNull()
            .observe(this, Observer {
                userFirstName = it.firstName.nullToEmpty()
                userLastName = it.lastName.nullToEmpty()
                val imageUrl = it.avatarUrl.nullToEmpty()
                rootView.firstName.setText(userFirstName)
                rootView.lastName.setText(userLastName)
                if (!imageUrl.isEmpty()) {
                    val drawable = requireDrawable(requireContext(), R.drawable.ic_account_circle_grey)
                    Picasso.get()
                        .load(imageUrl)
                        .placeholder(drawable)
                        .transform(CircleTransform())
                        .into(rootView.profilePhoto)
                }
            })
profileViewModel.fetchProfile()
```
The reason why I delete it and I think it's not useful to use the profileViewModel in EditProfileFragment is because:
1. I don't think it is wiseful to .fetchProfile() while it is already loaded one time in the ProfileFragment, so it will save the resource to just pass the data right from ProfileFragment.
2. The profileViewModel here observe the changes on user data, which will not be changed unless we update the user data -> which bring us back to ProfileFragment. The profileViewModel in here doesn't observe and retain the change on EditText fields.

That's why I make this PR. Maybe you misunderstood my method because I wasn't clear on the word "data loss" here, by "data" i mean the text typed or changed in the EditText, not the data of user (firstname, lastname)

iamareebjamal:
In viemodel, simply don't fetch profile if it is already loaded. Secondly, submit the text changes to the view model which will store it in the live data and then when you rotate it, it'll be available to you

anhanh11001:
Thanks for your comment,
In the change I made here, basically, I don't fetch the profile anymore, but I just pass the data fetched before from ProfileFragment to EditProfileFragment by bundle and key strings.

About the profileViewModel in the EditProfileFragment, I still think it's unnecessary because, by default, the editText already retain its information when it is given an id. So it doesn't make sense to me to do extra works to set up the ViewModel and live data, which tbh I think is also the cause of this issue #1188 as it doesn't observe the change of the text inside EditText but the change of user information.

Please take a look closely when you have the time!

PR #1204, opened 2019-02-28
refactor: Migrate EventDetailsFragment to use SafeArgs

### LINKED ISSUE #1198: Migrate to safe args in EventFragment and EventDetailsFragment
Currently `EventFragment` and `EventDetailsFragment` use raw bundles to pass data around, and do not take advantage of the type safe args gradle plugin released with the Navigation Architecture Component.

**Additional context**
Type safe args would ensure type safety when retrieving Fragment arguments from their bundles.

- [x] Yes

haroldadmin:
Fixes #1198 

Changes: Migrates the EventDetailsFragment to use Safe args instead of working with raw bundles. Arguments have been added to the navigation graph XML file, along with a more idiomatic Kotlin approach to creating argument bundles for related fragments. The delegate function navArgs has been used to retrieve the arguments for this fragment

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -152,8 +152,8 @@ dependencies {
     implementation 'com.journeyapps:zxing-android-embedded:3.6.0'
 
     //Navigation
-    implementation 'android.arch.navigation:navigation-fragment:1.0.0-rc02'
-    implementation 'android.arch.navigation:navigation-ui:1.0.0-rc02'
+    implementation 'android.arch.navigation:navigation-fragment-ktx:1.0.0-rc02'
+    implementation 'android.arch.navigation:navigation-ui-ktx:1.0.0-rc02'
 
     testImplementation 'junit:junit:4.12'
     testImplementation "io.mockk:mockk:1.9.1"
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -18,6 +18,7 @@ import androidx.core.view.isGone
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.navigation.Navigation.findNavController
+import androidx.navigation.fragment.navArgs
 import com.google.android.material.snackbar.Snackbar
 import com.squareup.picasso.Picasso
 import kotlinx.android.synthetic.main.content_event.aboutEventContainer
@@ -48,8 +49,7 @@ import org.fossasia.openevent.general.R
 import org.fossasia.openevent.general.event.EventUtils.loadMapUrl
 import org.fossasia.openevent.general.event.topic.SimilarEventsFragment
 import org.fossasia.openevent.general.social.SocialLinksFragment
-import org.fossasia.openevent.general.ticket.CURRENCY
-import org.fossasia.openevent.general.ticket.TicketsFragment
+import org.fossasia.openevent.general.ticket.TicketsFragmentArgs
 import org.fossasia.openevent.general.utils.Utils.getAnimSlide
 import org.fossasia.openevent.general.utils.Utils.requireDrawable
 import org.fossasia.openevent.general.utils.extensions.nonNull
@@ -59,13 +59,14 @@ import org.koin.androidx.viewmodel.ext.android.viewModel
 import timber.log.Timber
 import java.util.Currency
 
-const val EVENT_ID = "EVENT_ID"
+const val EVENT_ID = "eventId"
 const val EVENT_TOPIC_ID = "EVENT_TOPIC_ID"
 
 class EventDetailsFragment : Fragment() {
     private val eventViewModel by viewModel<EventDetailsViewModel>()
+    private val safeArgs: EventDetailsFragmentArgs by navArgs()
+
     private lateinit var rootView: View
-    private var eventId: Long = -1
     private var eventTopicId: Long? = null
     private lateinit var eventShare: Event
     private var currency: String? = null
@@ -77,11 +78,6 @@ class EventDetailsFragment : Fragment() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        val bundle = this.arguments
-        if (bundle != null) {
-            eventId = bundle.getLong(EVENT_ID, -1)
-        }
-
         eventViewModel.event
             .nonNull()
             .observe(this, Observer {
@@ -99,7 +95,7 @@ class EventDetailsFragment : Fragment() {
                 }
                 runOnce = false
 
-                Timber.d("Fetched events of id %d", eventId)
+                Timber.d("Fetched events of id %d", safeArgs.eventId)
             })
     }
 
@@ -126,7 +122,7 @@ class EventDetailsFragment : Fragment() {
                 Snackbar.make(rootView.eventCoordinatorLayout, it, Snackbar.LENGTH_LONG).show()
             })
 
-        eventViewModel.loadEvent(eventId)
+        eventViewModel.loadEvent(safeArgs.eventId)
 
         // Set toolbar title to event name
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
@@ -191,7 +187,7 @@ class EventDetailsFragment : Fragment() {
         // About event on-click
         val aboutEventOnClickListener = View.OnClickListener {
             val bundle = Bundle()
-            bundle.putLong(EVENT_ID, eventId)
+            bundle.putLong(EVENT_ID, safeArgs.eventId)
             findNavController(rootView).navigate(R.id.aboutEventFragment, bundle, getAnimSlide())
         }
 
@@ -279,7 +275,7 @@ class EventDetailsFragment : Fragment() {
                 return true
             }
             R.id.favorite_event -> {
-                eventViewModel.setFavorite(eventId, !(eventShare.favorite))
+                eventViewModel.setFavorite(safeArgs.eventId, !(eventShare.favorite))
                 if (eventShare.favorite) {
                     setFavoriteIcon(R.drawable.ic_baseline_favorite_border_white)
                 } else {
@@ -334,20 +330,21 @@ class EventDetailsFragment : Fragment() {
     }
 
     private fun loadTicketFragment() {
-        // Initialise Ticket Fragment
-        val ticketFragment = TicketsFragment()
-        val bundle = Bundle()
-        bundle.putLong("EVENT_ID", eventId)
-        bundle.putString(CURRENCY, currency)
-        ticketFragment.arguments = bundle
-        findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
+        TicketsFragmentArgs.Builder()
+            .setEventId(safeArgs.eventId)
+            .setCurrency(currency)
+            .build()
+            .toBundle()
+            .also { bundle ->
+                findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
+            }
     }
 
     private fun loadSocialLinksFragment() {
         // Initialise SocialLinks Fragment
         val socialLinksFragemnt = SocialLinksFragment()
         val bundle = Bundle()
-        bundle.putLong("EVENT_ID", eventId)
+        bundle.putLong(EVENT_ID, safeArgs.eventId)
         socialLinksFragemnt.arguments = bundle
         val transaction = childFragmentManager.beginTransaction()
         transaction.add(R.id.frameContainerSocial, socialLinksFragemnt).commit()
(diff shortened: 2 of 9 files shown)


haroldadmin:
> Build is failing. Please do not use wildcard imports and revert unnecessary changes.

On it. Will update the PR in a few minutes.

haroldadmin:
@liveHarshit I've fixed the build issue. Please take a look and review when you can.

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -314,32 +310,42 @@ class EventDetailsFragment : Fragment() {
...
-        bundle.putLong("EVENT_ID", eventId)
-        bundle.putString(CURRENCY, currency)
-        ticketFragment.arguments = bundle
-        findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
+        Bundle().apply {
+            putLong(EVENT_ID, safeArgs.eventid)
+            putString(CURRENCY, currency)
+        }.also { bundle ->
+            ticketFragment.arguments = bundle
+            findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
```
liveHarshit: Why are you passing bundle if safe args are used? Get the data from safe args generated class, no need to pass it while navigation.

app/build.gradle
```diff
@@ -152,8 +152,8 @@ dependencies {
     implementation 'com.journeyapps:zxing-android-embedded:3.6.0'
 
     //Navigation
-    implementation 'android.arch.navigation:navigation-fragment:1.0.0-rc01'
-    implementation 'android.arch.navigation:navigation-ui:1.0.0-rc02'
+    implementation 'android.arch.navigation:navigation-fragment-ktx:1.0.0-rc01'
+    implementation 'android.arch.navigation:navigation-ui-ktx:1.0.0-rc02'
```
liveHarshit: Safe args plugin is already added. Why this?

haroldadmin: Using the KTX version gives me access to the `navArgs`delegate. It's a lot cleaner to do it like this, and this way the arguments are evaluated lazily.

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
haroldadmin: Tickets Fragment does not use SafeArgs, so there is no generated args class for TicketsFragment. Passing a bundle is the only way to do it right now. I have opened [another issue](https://github.com/fossasia/open-event-android/issues/1205) to migrate the entire project to use safe args. We won't have to use bundles then.

Also, I think to use safe args directly with the navigate call we have to use Navigation AAC's "Action" feature. We don't use it currently in this app, so we have to rely on passing bundles only. SafeArgs does help with parsing those bundles, though. So I guess we are gaining some benefit here, at least.

app/build.gradle
liveHarshit: OK

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
liveHarshit: > Tickets Fragment does not use SafeArgs, so there is no generated args class for TicketsFragment.

Then do it for both in proper way. No need to pass bundles while navigation with safe args.

liveHarshit: > I think to use safe args directly with the navigate call we have to use Navigation AAC's "Action" feature. We don't use it currently in this app, so we have to rely on passing bundles only. SafeArgs does help with parsing those bundles, though. So I guess we are gaining some benefit here, at least.

@iamareebjamal Your views?

haroldadmin: > > Tickets Fragment does not use SafeArgs, so there is no generated args class for TicketsFragment.
> 
> Then do it for both in proper way. No need to pass bundles while navigation with safe args.

I'm not sure if that's possible. We can only use the generated Args class as argument bundle to the `navigate()` call if we are using Actions from Nav AAC. Currently we just have a nav graph with fragment destinations only. We don't use actions to define paths between them currently.

It's better to convert the graph to use actions first. [This page](https://developer.android.com/topic/libraries/architecture/navigation/navigation-implementing#Connect-destinations) shows how it can be done.

I currently use Nav AAC with actions and safe Args in one of my own projects, and I think navigating via actions is a lot better than just passing the ID resource of the destination fragment.

haroldadmin: Actions also help us visualize the possible app flows in the Nav editor in Android Studio. If you want, I can open a separate issue for it and work on it. But before we do that, i think it's better if we first define the args for all fragments in the app in nav_graph.xml file first. I'm referring to [this issue for it](https://github.com/fossasia/open-event-android/issues/1205)

iamareebjamal: I don't think it's true that you HAVE to use actions to use safe args. I may be wrong, but just arguments to destinations should generate helper classes to create and parse bundles into safe args

haroldadmin: Yes, just arguments for destinations do generate helper classes, and you can use them to generate arguments safely. However, if you are not using actions you will have to convert them to bundle in the end to pass them to the `navigate()` call.

For example, to navigate from EventsFragment to EventDetailsFragment, you can generate arguments this way: 
```
EventDetailsFragmentArgs.Builder()
                    .setEVENTID(eventID)
                    .build()
                    .toBundle()
                    .also { bundle ->
                        findNavController(rootView).navigate(R.id.eventDetailsFragment, bundle, getAnimFade())
                    }
```
Here I generate args using the Helper class, and then convert them to a bundle in the end.

If we were using Actions, though, we could do it this way: (this is a code snippet from one of my apps)

```
MovieDetailsFragmentDirections.actionMovieDetailsFragmentToActorDetailsFragment()
                .apply {
                    actorIdArg = id
                    transitionNameArg = transitionName
                }
                .also { action ->
                        findNavController().navigate(action)
                }
```

I have added arguments to the nav graph for EventDetailsFragment, but not to TicketsFragment. Hence there is no helper class for TicketsFragment, and I have to resort to using regular bundles.

I have not added args to nav graph for tickets fragment or others because I thought it was out of scope for this PR. I have opened a separate issue to migrate the entire project to use safe args [here](https://github.com/fossasia/open-event-android/issues/1205). Once this migration is complete, we can eventually transition the app to use Actions.

haroldadmin:
EDIT: Updated PR to rebase on latest upstream dev branch

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
liveHarshit: Follow https://codelabs.developers.google.com/codelabs/android-navigation/#7 and see how it is done for search date while navigating to search from search time fragment.

liveHarshit: Yes, using nav actions is a good idea but it is complicated because here we are not navigation to one fragment from one destination only.

haroldadmin: @liveHarshit I agree with you on this. Actions are suitable only when your screen hierarchy is well defined.

I am currently working on updating the TicketsFragment to use SafeArgs, so that we can finally settle this problem

haroldadmin: I've pushed some new changes. TicketsFragment now uses SafeArgs too. I have changed the EventDetailsFragment to use the Builder method of the generated Args class of TicketsFragment.

Hope that resolves this issue 😊

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -312,20 +308,21 @@ class EventDetailsFragment : Fragment() {
...
-        bundle.putString(CURRENCY, currency)
-        ticketFragment.arguments = bundle
-        findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
+        TicketsFragmentArgs.Builder()
+            .setEVENTID(safeArgs.eventid)
+            .setCURRENCY(currency)
+            .build()
+            .toBundle()
+            .also { bundle ->
+                findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
```
liveHarshit: Why are passing bundle if you are getting event id from safe args in ticket fragment?

haroldadmin: Because `findNavController().navigate()` call does not accept safe-args directly. You can not pass an instance of `TicketFragmentArgs` directly to it. You have to create safe args using the builder, convert that builder to a bundle, and then pass that bundle to the navigate call.

Here's the error I get when I try to pass an instance of TicketFragmentArgs directly without converting it to a bundle first:
screenshot 2019-03-01 at 3 15 05 pm


If you want to pass arguments without using a bundle, then you have to encapsulate them in a NavDirections object. Our project does not use Actions from the Nav AAC, so we don't have generated NavDirections classes for any fragments.

I feel like I am unable to understand what seems to be the issue here. I apologize if I am still wrong, can you please explain in detail to me the method you are suggesting?

liveHarshit: Just use `.build().toBundle()`.

liveHarshit: Can't we get eventId directly from the generated class in tickets fragment?

haroldadmin: > Just use `.build().toBundle()`.

That is kind of exactly what I am doing right now, except that instead of storing the value returned by the `toBundle()` call in a variable, I use Kotlin's `also` function to directly pass that value to the `navigate` call.

> Can't we get eventId directly from the generated class in tickets fragment?

That's exactly what we are doing right now. The `by navArgs()` delegate takes the received arguments for the fragment and type-casts them to the corresponding generated Args class (which is TicketsFragmentArgs in this case). 

SafeArgs are not a replacement for bundles, they are just a way to introduce type safety to them. Under the hood they still use bundles, but they help in verification of them to ensure that requested arguments are present inside them.

haroldadmin:
> Okay, It is right, we need to pass bundle after setting data to safe args. I apologize for the inconvenience.

No worries! The discussion helped me understand the Nav AAC better. Please give a final review to the PR and merge if everything looks okay 😊

haroldadmin:
Rebased on latest upstream/development branch changes.

haroldadmin:
@liveHarshit @iamareebjamal Can this be merged now? I am basing the branch to migrate entire project to safe args on this one, and if i submite that PR before this one gets merged then there will be some merge conflicts to solve.

liveHarshit:
Please wait for @iamareebjamal's reviews. I don't have the rights.

haroldadmin:
@nikit19 @iamareebjamal I've updated the PR to include the latest changes in upstream development branch. Please review when you can, and merge if everything looks good.

haroldadmin:
@iamareebjamal Can you please review this PR?

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -334,20 +330,21 @@ class EventDetailsFragment : Fragment() {
...
-        // Initialise Ticket Fragment
-        val ticketFragment = TicketsFragment()
-        val bundle = Bundle()
-        bundle.putLong("EVENT_ID", eventId)
-        bundle.putString(CURRENCY, currency)
-        ticketFragment.arguments = bundle
-        findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
+        TicketsFragmentArgs.Builder()
+            .setEVENTID(safeArgs.eventid)
+            .setCURRENCY(currency)
```
iamareebjamal: Can we change this to camel case

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -334,20 +330,21 @@ class EventDetailsFragment : Fragment() {
...
+                findNavController(rootView).navigate(R.id.ticketsFragment, bundle, getAnimSlide())
+            }
     }
 
     private fun loadSocialLinksFragment() {
         // Initialise SocialLinks Fragment
         val socialLinksFragemnt = SocialLinksFragment()
         val bundle = Bundle()
-        bundle.putLong("EVENT_ID", eventId)
+        bundle.putLong("EVENT_ID", safeArgs.eventid)
```
iamareebjamal: Change to constant

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
haroldadmin: Yes, I'll add this change right now.

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
haroldadmin: Okay, will update the PR in a few mins.

haroldadmin:
@iamareebjamal I have fixed all the problems you found in the codereview, and have updated the PR to rebase on current upstream dev branch.

haroldadmin:
@iamareebjamal Updated the PR again.

PR #1228, opened 2019-03-05
fix: Added Jackson Library

### LINKED ISSUE #1227: Add Jackson Library in Readme
**Describe the chore**
There is no mention of jackson library in readme.

- [x] Yes

somenath1435:
Fixes #1227 

Changes:Added Jackson Library in Readme with proper documentation link.


somenath1435:
@iamareebjamal But the library is important for the codebase. It should definitely be mentioned in Readme so that contributors can learn about it (Documentation attached)

iamareebjamal:
I think all libraries should be removed but Jackson is not something you need to learn to be able to code the app. There's no code of Jackson we're using anywhere. It's just an abstraction. We can just as easily replace it with gson or moshi

PR #1252, opened 2019-03-09
fix: wrong colors in Buttons in Login/Signup Fragment

### LINKED ISSUE #1249: Wrong colors in Buttons in Login/SingUp fragment
**Describe the chore**
The background color of login/signup fragment is wrong and looks bad.
Need to change it.

- [x] Yes

somenath1435:
Fixes #1249 

Changes: Changed the colors.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -22,7 +22,7 @@
     </style>
 
     <style name="AccentButton" parent="ThemeOverlay.AppCompat.Dark">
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -22,7 +22,7 @@
     </style>
 
     <style name="AccentButton" parent="ThemeOverlay.AppCompat.Dark">
```
iamareebjamal: Why are we even using this style?

PR #1257, opened 2019-03-10
fix: edit text retention in EditProfileFragment

### LINKED ISSUE #652: Name and Last Name are not retained on rotation in edit profile activity
**Describe the bug**
While Setting up the name and last name after logIn in the Edit Image activity the text entered is not retained on rotation

**To Reproduce**
1. Go to Profile.
2. Click on Edit Profile.
3. Enter First Name and last name but do not press update.
4. Rotate the device to landscape orientation.
4. See error your entered text is not saved.

**Expected behavior**

Text should be retained

**Screenshots**

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | OnePlus5 |
|Android Version|  Oreo 8.1.0 |

 Yes

anhanh11001:
Details:
I checked if the EditText already contains text -> if it is then it shouldn't reset the text when rotating

Fixes #652 

Screenshots for the change:
ezgif-1-3751bc4ad94a


iamareebjamal:
Please align your commit messages with the guideline

PR #1273, opened 2019-03-11
fix: stop reloading old image when rotating on EditProfileFragment

### LINKED ISSUE #1016: User Details and Image is lost in screen rotation

anhanh11001:
Details:
When the screen is rotated, if a new image is chosen, the URI of that image is saved on onSaveInstanceState and then reload when the view is created after rotating

Fixes #1016 (The part where EditText is not retained in 1016 issue is already resolved in other PR and merged)

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/ProfileViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/ProfileViewModel.kt
@@ -19,6 +19,7 @@ class ProfileViewModel(private val authService: AuthService) : ViewModel() {
     val user: LiveData<User> = mutableUser
     private val mutableError = SingleLiveEvent<String>()
     val error: LiveData<String> = mutableError
+    var avatarPicked = MutableLiveData<String>()
 
     fun isLoggedIn() = authService.isLoggedIn()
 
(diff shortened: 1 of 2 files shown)


iamareebjamal:
Use view model. No need for saved instance state

anhanh11001:
@iamareebjamal  I have used the viewModel, please take a look

app/src/main/java/org/fossasia/openevent/general/auth/ProfileViewModel.kt
```diff
@@ -19,6 +19,7 @@ class ProfileViewModel(private val authService: AuthService) : ViewModel() {
     val user: LiveData<User> = mutableUser
     private val mutableError = SingleLiveEvent<String>()
     val error: LiveData<String> = mutableError
+    var avatarPicked = MutableLiveData<String>()
```
iamareebjamal: Should be val

PR #1290, opened 2019-03-13
fix: display placeholder image for event

### LINKED ISSUE #1289: No placeholder image for events
**Describe the chore**

No placeholder image is being displayed for events having no image and instead a white blank space is present.
The default placeholder image should be used in such a case

New Issue

It should look something like this
New PR

- [x] Yes
- [ ] No
- Other:

addiegupta:
Fixes #1289.

Changes: 
Instead of displaying the placeholder image only in the layout editor by using the namespace "tools" , the android namespace is used to display the placeholder image by default
Additional changes:
- Removed old placeholder image that is not being used
- Collapsed empty tag for FrameLayout in content_event.xml

Screenshots for the change:
New PR

Edit: As requested by @iamareebjamal , placeholder image has been updated

PR 1290


iamareebjamal:
This should not be the placeholder image. Take the one used in open event orga app

addiegupta:
I have updated the placeholder image @iamareebjamal

aman210697:
> I have updated the placeholder image @iamareebjamal

How about using Picasso to set placeholder?

iamareebjamal:
Yes it should be used

addiegupta:
@iamareebjamal using Picasso to set the placeholder results in some imageview being empty while in some the placholder is applied. Should I add Picasso placeholder and set src in XML as well?

iamareebjamal:
Yes

addiegupta:
@iamareebjamal  updated

PR #1293, opened 2019-03-13
fix: Retain ticket details' visibility after screen rotation

### LINKED ISSUE #1286: Hide/View is not retained after rotating screen
**Describe the bug**
When you are opening the detail section on price by (view), it's closed after rotating devices. So the current state is not retained.

**Screenshots**

Yes

anhanh11001:
Details:
Ticket Details Visibility is subscribed to the View Model as a boolean so that it survives after screen rotation

Fixes #1286 

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
@@ -57,6 +57,7 @@ class AttendeeViewModel(
     private val mutableQtyList = MutableLiveData<ArrayList<Int>>()
     val qtyList: LiveData<ArrayList<Int>> = mutableQtyList
     val paymentCompleted = MutableLiveData<Boolean>()
+    val ticketDetailsVisibility = MutableLiveData<Boolean>()
     private val mutableTickets = MutableLiveData<MutableList<Ticket>>()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -143,6 +144,14 @@ class AttendeeViewModel(
         )
     }
 
+    fun updateTicketDetailsVisibility() {
+        if (ticketDetailsVisibility.value == null) {
+            ticketDetailsVisibility.value = false
+        } else {
+            ticketDetailsVisibility.value = !ticketDetailsVisibility.value!!
```
liveHarshit: Please do not use non-null assertion.

anhanh11001: I will fix this. And some of my PRs keep having build fail because UndocumentedPublicFunction error, so how should I avoid this or does that mean I have to write some documentations/comments for that ?

liveHarshit: I think build will not fail for UndocumentedPublicFunction. Codacy will fail, which is okay.

PR #1294, opened 2019-03-13
feat: Add remove favorite event animation

### LINKED ISSUE #1285: Add animation on removing favorite event
Add some kind of destroy animation on removing favorite event, not just plain disappearing.

**Screenshots**

Yes

anhanh11001:
Details:
There is a problem in the FavoriteFragment: The way it updates the favorite event list is wrong. Basically, when an event is clicked on the favorite icon, it is not removed from the adapter events' list. But as we call `favoriteEventViewModel.setFavorite(event.id, false)`, it is removed by the viewModel. And then the viewModel.events change so it notifies the code below inside FavoriteFragment:
```
favoriteEventViewModel.events
            .nonNull()
            .observe(this, Observer {
                favoriteEventsRecyclerAdapter.addAll(it)
                favoriteEventsRecyclerAdapter.notifyDataSetChanged()
                showEmptyMessage(favoriteEventsRecyclerAdapter.itemCount)

                Timber.d("Fetched events of size %s", favoriteEventsRecyclerAdapter.itemCount)
            })
```
As the result, the whole event list is fetched again.

I fixed this by making sure the viewModel only fetch the list once, remove the favorite event properly from the adapter and use notifyItemRemoved(), not notifyItemChanged() to make the animations.

Fixes #1285 

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
@@ -34,6 +34,7 @@ class FavoriteFragment : Fragment() {
     private val favoriteEventsRecyclerAdapter: FavoriteEventsRecyclerAdapter = FavoriteEventsRecyclerAdapter()
     private val favoriteEventViewModel by viewModel<FavoriteEventsViewModel>()
     private lateinit var rootView: View
+    private var favoriteListFetched: Boolean = false
 
     override fun onCreateView(
         inflater: LayoutInflater,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
```diff
@@ -34,6 +34,7 @@ class FavoriteFragment : Fragment() {
     private val favoriteEventsRecyclerAdapter: FavoriteEventsRecyclerAdapter = FavoriteEventsRecyclerAdapter()
     private val favoriteEventViewModel by viewModel<FavoriteEventsViewModel>()
     private lateinit var rootView: View
+    private var favoriteListFetched: Boolean = false
```
liveHarshit: No need to declare `Boolean`. You can directly take it false.

iamareebjamal:
Even better would be using ListAdapter which would have internally used DiffUtils

iamareebjamal:
Great work on your commit descriptions. But please always add # symbol before issue number so github makes it a hyperlink and links to the issue and rather than writing in unparseable 'This resolves issue 1234' which github does not understand, simply write `Fixes #1234` or `Closes #1234` and github will understand that this PR closes this particular issue. That's why you see it written as such in the PR template

Issue #1292, opened 2019-03-13
EditProfile Alert Dialog shown on Up button press but not on back press


addiegupta:
**Describe the chore**

An AlertDialog is displayed in the EditProfile Fragment when some changes have been made to the profile and the up button is pressed. This dialog warns the user of unsaved changes. However this dialog is not displayed when the back button is pressed.

The back button should also display the dialog.

GIF showing the issue

<table>
<tr>
<td><img src = "https://user-images.githubusercontent.com/22665789/54286368-b93c7180-45c9-11e9-8880-ee819b5bd693.gif" height ="533" width ="300"/>
</td>
</tr>
</table

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

PR #1307, opened 2019-03-14
fix: EditProfile discard changes dialog shown on back press

addiegupta:
Fixes Issue #1292. The behaviour for up button has been replicated for the back button. So now if changes are made in the EditProfileFragment and the back button or the up button is pressed, an AlertDialog is displayed which warns the user of the unsaved changes

Changes: 
The MainActivity class has been modified to handle the backPress for the EditProfileFragment. It in turn calls a method in EditProfileFragment which handles the backPress. When the back button is finally meant to be pressed from EditProfileFragment to go back, then super.onBackPressed is called using a new method.

Please suggest a better method to implement this if any.

Screenshots for the change:
pr for 1292


iamareebjamal:
The solution presented is very brittle and can fail in many ways

addiegupta:
> The solution presented is very brittle and can fail in many ways

Can you guide me towards an efficient solution?

iamareebjamal:
Why you haven't just overriden onBackPressed in the fragment?

addiegupta:
> Why you haven't just overriden onBackPressed in the fragment?

As far as I checked onBackPressed cannot be overriden in fragments. That is why I took this route.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -68,10 +69,23 @@ class MainActivity : AppCompatActivity() {
...
+            }
             else -> super.onBackPressed()
         }
     }
 
+    /**
+     * Called by EditProfileFragment to go to previous fragment
+     */
+    fun onSuperBackPressed() {
+        super.onBackPressed()
```
iamareebjamal: Useless as it's just calling onBackPressed. OnBackPressed can be called directly

addiegupta: I use this method to call the direct onBackPressed from the EditProfileFragment. If i would have called the onBackPressed then it would have called the overriden method which would again display the AlertDialog

iamareebjamal: Ok

addiegupta: Any changes I should make?

addiegupta:
@iamareebjamal I have updated the branch. Please review

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -68,10 +69,23 @@ class MainActivity : AppCompatActivity() {
             }
             R.id.orderCompletedFragment -> navController.popBackStack(R.id.eventDetailsFragment, false)
             R.id.welcomeFragment -> finish()
+            R.id.editProfileFragment -> {
+
+                // Calls the handleBackPress method in EditProfileFragment
+                val hostFragment = supportFragmentManager.findFragmentById(R.id.frameContainer) as? NavHostFragment
+                (hostFragment?.childFragmentManager?.fragments?.get(0) as? EditProfileFragment)?.handleBackPress()
```
liveHarshit: Too complex.

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
```diff
@@ -110,7 +111,7 @@ class EditProfileFragment : Fragment() {
             .observe(this, Observer {
                 Snackbar.make(rootView.editProfileCoordinatorLayout, it, Snackbar.LENGTH_LONG).show()
                 if (it == USER_UPDATED) {
-                    activity?.onBackPressed()
+                    (activity as? MainActivity)?.onSuperBackPressed()
```
liveHarshit: Do not cast.

addiegupta: How do I use this method from MainActivity then without casting?

liveHarshit: Use smart cast instead.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
addiegupta: Should I split these up into more lines of code or do you mean some other way should be used? Because I could not find any other way

addiegupta:
> You need to check if edit text and image field are clicked, then only show an alert dialog.

This is done in the following code

`if (!avatarUpdated && rootView.lastName.text.toString() == userLastName &&
            rootView.firstName.text.toString() == userFirstName)`

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
```diff
@@ -110,7 +111,7 @@ class EditProfileFragment : Fragment() {
             .observe(this, Observer {
                 Snackbar.make(rootView.editProfileCoordinatorLayout, it, Snackbar.LENGTH_LONG).show()
                 if (it == USER_UPDATED) {
-                    activity?.onBackPressed()
+                    activity?.let { act -> if (act is MainActivity) act.onSuperBackPressed() }
                 }
```
addiegupta: @liveHarshit  I have done it this way instead of using `if(activity is MainActivity) activity.onSuperBackPressed()` because the following error occurred for this method

`Smart cast to 'MainActivity' is impossible, because 'activity' is a property that has open or custom getter`

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
addiegupta: Changes made

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
liveHarshit: Then use a value - `val thisActivity = activity` then cast for thisActivity.

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
```diff
@@ -219,6 +205,29 @@ class EditProfileFragment : Fragment() {
         }
     }
 
+    /**
+     * Handles back press when up button or back button is pressed
+     */
+    fun handleBackPress() {
+        val activity = activity as? MainActivity
```
liveHarshit: Don't cast like that.

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
addiegupta: okay! I tried that also but thought this was cleaner. will implement this

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
addiegupta: okay

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
liveHarshit: You just need to call the method in edit profile fragment. Can't you do it by taking the fragment as object?

addiegupta: How do I get a reference to the Fragment instance?

addiegupta:
@iamareebjamal @liveHarshit any other changes that I should make in this?

liveHarshit:
@iamareebjamal Please review.

PR #1327, opened 2019-03-17
Add Support for Anko Layouts

### LINKED ISSUE #1320: Anko Layouts
Use Anko DSL Layouts instead of xml especially for recyclerview items 
Why Anko Layouts ?
https://android.jlelse.eu/400-faster-layouts-with-anko-da17f32c45dd
- [x] Yes
- [ ] No
- Other:

aggarwalpulkit596:
Fixes #1320 

Changes: Converted Item Card Event to Anko DSL

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/ItemCardEventsUi.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/ItemCardEventsUi.kt
@@ -1,0 +1,132 @@
+package org.fossasia.openevent.general.event
+
+import android.annotation.SuppressLint
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.constraintlayout.widget.ConstraintLayout
+import org.fossasia.openevent.general.R
+import org.jetbrains.anko.*
+import org.jetbrains.anko.cardview.v7.cardView
+import org.jetbrains.anko.constraint.layout.constraintLayout
+import org.jetbrains.anko.design.floatingActionButton
+
+
+class ItemCardEventsUi : AnkoComponent<ViewGroup> {
+
+    companion object {
+        val eventImage = 1
+        val date = 2
+        val month = 3
+        val locationName = 4
+        val eventName = 5
+        val shareFab = 6
+        val favFab = 7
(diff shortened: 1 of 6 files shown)


aggarwalpulkit596:
@iamareebjamal review this you might have to run this on an actual device to see the difference in results and also how can i test it with actual result currently i was able to test with sample data only

app/src/main/java/org/fossasia/openevent/general/event/ItemCardEventsUi.kt
```diff
@@ -0,0 +1,132 @@
...
+
+import android.annotation.SuppressLint
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.constraintlayout.widget.ConstraintLayout
+import org.fossasia.openevent.general.R
+import org.jetbrains.anko.*
```
ShridharGoel: Build is failing because of wildcard import

aggarwalpulkit596:
Thanks for pointing it out but i already know that and i am waiting for it review first and PR is incomplete as of now

aggarwalpulkit596:
@iamareebjamal please review this

iamareebjamal:
Yeah I don't think it looks good or maintainable. Plus, I'm planning on using data binding later which we can't do here

aggarwalpulkit596:
We can use databinding as well checkout this video https://youtu.be/dYskPdOSvo4

iamareebjamal:
You can see the example in the video itself how verbose and unpleasant it is. People won't want to use databinding if they had to use it like that. Also, you can't mix anko layouts with normal layouts AFAIK. I want to include a layout and bind parameters in the XML, I can't do that

aggarwalpulkit596:
We can use include as well but nm I'll close this PR

iamareebjamal:
I vote against anko layouts because where anko layouts shine is when you have heavy dynamic layouts or want to remove overhead of layout inflation in case of simple layouts, etc. We don't have dynamic layouts, nor is layout inflation a bottleneck for us

iamareebjamal:
You can add anko layouts in new layouts we add for simple layouts but I don't think we should convert existing layouts

PR #1332, opened 2019-03-17
fix: refresh events on location change

### LINKED ISSUE #1331: Events not getting refreshed on changing location
**Describe the bug**

The events do not get refreshed on their own when the location is changed.

**To Reproduce**

The events should get updated as the location is no longer the same

**Screenshots**

Issue 1331

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Nokia 6.1 Plus|
|Android Version|Pie 9.0   |

**Additional context**

- [x] Yes
- [ ] No
- Other:

addiegupta:
Fixes #1331 

Changes: 
- In the EventsViewModel, the method that updates the savedLocation has been modified to now return a Boolean value: true if the location has changed, false otherwise. 

- This boolean is used in onCreateView() method(which is called on returning from SearchLocationFragment). If the location has changed, then the events are reloaded. Otherwise the initial fetch is made. 

- Note that if there is no change in location then this initial fetch is not repeated due to following if statement in the EventsViewModel (mutableEvents.value is not null)
`if (loadingTag == RELOADING_EVENTS || (loadingTag == INITIAL_FETCHING_EVENTS && mutableEvents.value == null)) `
       
Screenshots for the change:
PR 1332

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
@@ -133,9 +133,12 @@ class EventsFragment : Fragment() {
                 rootView.swiperefresh.isRefreshing = it
             })
 
-        eventsViewModel.loadLocation()
+        if (eventsViewModel.loadLocation()) {
+            eventsViewModel.loadLocationEvents(RELOADING_EVENTS)
+        } else {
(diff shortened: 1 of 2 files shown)


liveHarshit:
IMO, Unnecessary feature.

addiegupta:
This issue has apparently been fixed by #1350

PR #1350, opened 2019-03-19
fix: making search in EventsFragment works again

### LINKED ISSUE #1349: Search doesn't work in EventsFragment
**Describe the bug**
Needs to reload in order for searching in EventsFragment works. This is due to my PR #1296, I know the cause of the problem so I will fix it

**Screenshots**

Yes

anhanh11001:
**Details:**
Add a String to store the last search so that if a new search come, it reloads events again
Fixes #1349

**Screenshots for the change:**
Search works again, doesn't need to swipe down to reload

When rotating screen, the events are still retained

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
@@ -24,7 +24,7 @@ class EventsViewModel(private val eventService: EventService, private val prefer
     val error: LiveData<String> = mutableError
     private val mutableShowShimmerEvents = MutableLiveData<Boolean>()
     val showShimmerEvents: LiveData<Boolean> = mutableShowShimmerEvents
-    private var lastSearch = String()
+    private var lastSearch = ""
 
     var savedLocation: String? = null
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -24,6 +24,7 @@ class EventsViewModel(private val eventService: EventService, private val prefer
     val error: LiveData<String> = mutableError
     private val mutableShowShimmerEvents = MutableLiveData<Boolean>()
     val showShimmerEvents: LiveData<Boolean> = mutableShowShimmerEvents
+    private val mutableLastSearch = MutableLiveData<String>()
```
liveHarshit: Why mutable String? Can't you use a simple String instead?

anhanh11001: You are right, it still works fine with String so I'm sending a new commit now

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -24,6 +24,7 @@ class EventsViewModel(private val eventService: EventService, private val prefer
     val error: LiveData<String> = mutableError
     private val mutableShowShimmerEvents = MutableLiveData<Boolean>()
     val showShimmerEvents: LiveData<Boolean> = mutableShowShimmerEvents
+    private var lastSearch = String()
```
iamareebjamal: Just use  ""

PR #1351, opened 2019-03-19
fix: Retain search results after rotating screens

### LINKED ISSUE #1347: Unnecessary reloading search result
**Describe the bug**
In Search Results Fragment, it keeps reloading search results when rotating devices

**Expected behavior**
That fragment should keep the search results in ViewModel and load it back

**Screenshots**

Yes

anhanh11001:
Details:
If there are already search results, they are reused, and the search query is not called again in the view model

Fixes #1347

Screenshots for the change:


liveHarshit:
Please add GIF.

anhanh11001:
@liveHarshit I have already added one below "Screenshots for the change:"...

PR #1374, opened 2019-03-22
refactor: Migrate project to use viewLifecycleOwner in fragments to observe livedata wherever it makes sense 

### LINKED ISSUE #1326: refactor: Migrate project to use viewLifecycleOwner in fragments to observe livedata wherever it makes sense
**Describe the bug**
Currently all of our fragments use the fragment as the lifecycle owner when subscribing to LiveData objects. This is problematic because:
* Subscribe to livedata in onCreateView/onViewCreated/onActivityCreated using fragment as the lifecycle owner
* This livedata will automatically unsubscribe when the fragment is torn down, as is the nature of livedata.
* When fragment is `replaced` with another one, the view hierarchy will be destroyed but the fragment will not be.
* Return to this fragment: LiveData is subscribed to again because onCreateView/onViewCreated/onActivityCreated is called again, without unsubscribing to the last subscription.
* Subscribing to ViewModel in methods called before these three can result in NPE because the ViewModel may not have been initialized by then.

[This article explains this problem very well](https://medium.com/@BladeCoder/architecture-components-pitfalls-part-1-9300dd969808)

Google added a solution for this in the form a `viewLifecycleOwner` property, which relates to the view lifecycle of the fragment instead of the actual fragment lifecycle. This means that when the view hierarchy is destroyed, any live data subscriptions will be unsubscribed too.

**Additional context**
There is excellent discussion regarding this problem on this Github issue: https://github.com/googlesamples/android-architecture-components/issues/47

- [x] Yes

haroldadmin:
Fixes #1326 

Changes: Change every fragment in the app that uses LiveData in some capacity to use the `viewLifecycleOwner` property wherever it seems appropriate. To decide where it makes sense to change the Lifecycle owner in a live data subscription, I looked at what actions are being invoked in the observer. If they are purely UI related events, I have changed it to `viewLifecycleOwner`.

* I have tried to make as little changes as possible. Wherever a LiveData object was being observed solely for the purpose of controlling UI events (Snackbars, errors, etc), I have changed the Observer to use the fragment's view lifecycle instead.

* I have tested the app after each commit to ensure nothing is being broken because of these changes.

* In some places where UI events were being controlled by a live data observable but the subscription was being made in the Fragment's `onCreate()` method, I have moved the observer to `onCreateView()`.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/about/AboutEventFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/about/AboutEventFragment.kt
@@ -34,16 +34,6 @@ class AboutEventFragment : Fragment(), AppBarLayout.OnOffsetChangedListener {
     private var title: String = ""
     private val safeArgs: AboutEventFragmentArgs by navArgs()
 
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-
-        aboutEventViewModel.event
-            .nonNull()
-            .observe(this, Observer {
-                loadEvent(it)
-            })
-    }
-
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         rootView = layoutInflater.inflate(R.layout.fragment_about_event, container, false)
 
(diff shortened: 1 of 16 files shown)


haroldadmin:
@iamareebjamal @liveHarshit I have made one commit per fragment in this PR to make it easier to review. It is ready to be squashed into one commit if you want. Please review when you can, and merge if everything looks good.

iamareebjamal:
One commit per view doesn't make it easier to review because we normally see all files changed at one but thanks

haroldadmin:
Oh. Right. Still new to reviewing PRs, didn't realize that you always see all files at once. Thanks for the quick merge!

Issue #1341, opened 2019-03-18
No Internet Layout displayed on rotation even when data is present in EventsFragment on connection loss


addiegupta:
**Describe the bug**

The No Connection available card is displayed in EventsFragment on rotation if connection is lost after loading events and rotation takes place.

**To Reproduce**

1. Go to EventsFragment and let events load
2. Switch off Internet connection
3. Rotate device
4. See error --> The No internet layout is displayed even when data was present

**Expected behavior**

2 implementations can be made out of which 1 will be chosen
1) Display the no internet layout along with the pre loaded data (I prefer this)
2) Display just the pre loaded data

If anyone has any suggestion on this please let me know!

**Screenshots**

![Issue 1341](https://user-images.githubusercontent.com/22665789/54553931-4e2fd800-49d9-11e9-84e4-dedee3873af7.gif)

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Nokia 6.1 Plus|
|Android Version|Pie 9.0   |

**Additional context**

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

AnEnigmaticBug:
This issue is there because the `EventsFragment` is checking the internet connectivity in the `onCreateView` method.

After loading the data, when you rotate the phone, the `Fragment`'s view is recreated and `onCreateView` is called again. Because the check happens in `onCreateView`, it ignores whether the data has already been loaded or not and simply displays the "No Internet" layout.

I see two possible solutions to this problem:
1. Check whether `events.value` is non null and non empty. If it is, display the events. Otherwise check for the internet connectivity.
2. Move to logic for checking internet connection into the `EventsViewModel`.

Here's my take on the two solutions:
1. Solution1 will be the easier of two and will take minimal changes to fix the bug. But it doesn't improve the code.
2. Solution2 will take slightly more effort but will improve the code by moving logic from `EventsFragment` to `EventsViewModel`.

liveHarshit:
IMO, using the view model is a better solution.

PR #1377, opened 2019-03-22
fix: show previously loaded events on rotation + internet loss

addiegupta:
Fixes: #1341

Changes:
1) EventsFragment in onCreateView, a check is made to display no internet card only if pre loaded events are not present and internet is not available
2) The boolean argument for showNoInternetScreen has been flipped for consistency i.e. when the argument is true, then NoInternet screen is displayed

Screenshots for the change:

PR 1377

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
@@ -131,19 +131,19 @@ class EventsFragment : Fragment() {
             findNavController(rootView).navigate(R.id.searchLocationFragment, null, getAnimSlide())
         }
 
-        showNoInternetScreen(isNetworkConnected(context))
+        showNoInternetScreen(!isNetworkConnected(context) && eventsViewModel.events.value.isNullOrEmpty())
 
         rootView.retry.setOnClickListener {
             val isNetworkConnected = isNetworkConnected(context)
             if (eventsViewModel.savedLocation != null && isNetworkConnected) {
                 eventsViewModel.loadLocationEvents(RELOADING_EVENTS)
             }
-            showNoInternetScreen(isNetworkConnected)
+            showNoInternetScreen(!isNetworkConnected)
         }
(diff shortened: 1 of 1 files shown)


addiegupta:
@iamareebjamal  please review

iamareebjamal:
Ask for peer reviews first

addiegupta:
> Ask for peer reviews first

okay sure

addiegupta:
@liveHarshit can you please review this?

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.general.event
 
+import android.content.Context
```
liveHarshit: No android import in view model.

addiegupta: Ill have to move the internet checking logic out of the ViewModel then. I did this because you suggested using the ViewModel in the discussions of the Issue #1341 for this PR. Should I proceed?

iamareebjamal: Yes wrap the check in another class with injected context

addiegupta: I have directly used the Util method to check for connection from EventsFragment passing the context

iamareebjamal: The ViewModel should be pure java code, no Android imports

addiegupta: @iamareebjamal sorry for the confusion. What I meant by most recent comment was referring to a new commit that I had pushed at that instant. In this, I have removed the Android import from ViewModel and have implemented checking of internet connectivity logic from the EventsFragment, passing the context from there only

Issue #1383, opened 2019-03-23
App crashing due to IllegalArgumentException 


addiegupta:
**Describe the bug**

The app is crashing on selecting Search from navigation bar

**To Reproduce**

The app should not crash 

**Logs**

FATAL EXCEPTION: main
    Process: com.eventyay.attendee, PID: 14541
    java.lang.IllegalArgumentException: Required argument "@string/saved_date" is missing and does not have an android:defaultValue
        at org.fossasia.openevent.general.search.SearchFragmentArgs.fromBundle(SearchFragmentArgs.java:34)
        at org.fossasia.openevent.general.search.SearchFragment$safeArgs$2.invoke(SearchFragment.kt:30)
        at org.fossasia.openevent.general.search.SearchFragment$safeArgs$2.invoke(SearchFragment.kt:24)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at org.fossasia.openevent.general.search.SearchFragment.getSafeArgs(Unknown Source:7)
        at org.fossasia.openevent.general.search.SearchFragment.onCreateView(SearchFragment.kt:59)
        at androidx.fragment.app.Fragment.performCreateView(Fragment.java:2539)
        at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:875)
        at androidx.fragment.app.FragmentManagerImpl.addAddedFragments(FragmentManagerImpl.java:2084)
        at androidx.fragment.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:1858)
        at androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManagerImpl.java:1814)
        at androidx.fragment.app.FragmentManagerImpl.execPendingActions(FragmentManagerImpl.java:1715)
        at androidx.fragment.app.FragmentManagerImpl$1.run(FragmentManagerImpl.java:143)
        at android.os.Handler.handleCallback(Handler.java:874)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:198)
        at android.app.ActivityThread.main(ActivityThread.java:6729)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

[...]

PR #1394, opened 2019-03-24
fix: app crash due to IllegalArgumentException in SearchFragment

addiegupta:
Fixes: #1383

Changes: 

- A default value of 'null' has been assigned to the argument 'saved_date'.
- This resolves the crash as the SearchFragmentArgs generated files throws an IllegalArgumentException if this argument
does not exist in the bundle with which it is initialised in its 'fromBundle' method. Now a null value is passed and hence the crash
does not occur.
- Additionally 2 hardcoded strings have been replaced with their resource reference equivalents

The code in SearchFragmentArgs that was throwing the exception is : 

`if (bundle.containsKey("@string/saved_date")) {`
     `// code that inserts it in arguments, redacted for simplicity`
    `} else {`
      `throw new IllegalArgumentException("Required argument \"@string/saved_date\" is missing and` `does not have an android:defaultValue");`
    `}`

After adding the `android:defaultValue="@null"` property to the `navigation.xml` file, this IllegalArgumentException statement appears to have gone from the generated file

Screenshots for the change:

PR 1383

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/SearchTimeFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/SearchTimeFragment.kt
@@ -105,7 +105,7 @@ class SearchTimeFragment : Fragment() {
     }
 
     private fun redirectToSearch(time: String) {
-        val args = SearchFragmentArgs.Builder(time).build().toBundle()
+        val args = SearchFragmentArgs.Builder().setStringSavedDate(time).build().toBundle()
         val navOptions = NavOptions.Builder().setPopUpTo(R.id.eventsFragment, false).build()
         Navigation.findNavController(rootView).navigate(R.id.searchFragment, args, navOptions)
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/openevent/general/search/SearchTimeFragment.kt
```diff
@@ -105,7 +105,7 @@ class SearchTimeFragment : Fragment() {
     }
 
     private fun redirectToSearch(time: String) {
-        val args = SearchFragmentArgs.Builder(time).build().toBundle()
+        val args = SearchFragmentArgs.Builder().setStringSavedDate(time).build().toBundle()
```
addiegupta: I made this change because for some reason that I could not find, the constructor for `SearchFragmentArgs.Builder` that takes a string as an argument appears to be missing from the generated file of `SearchFragmentArgs.java`

This is the missing constructor:

`  public Builder(@Nullable String StringSavedDate) {
      this.arguments.put("@string/saved_date", StringSavedDate);
    }
`

PR #1404, opened 2019-03-24
feat: improvise dirty build time

### LINKED ISSUE #1402: Improve dirty build time
**Describe the bug**
Current Dirty Build time for debug apk is around 3 min 41 sec (Macbook 2017 8gb ram 256gb ssd)
Build time can differ on different system,you can check it using
./gradlew assembleDebug --profile

Screenshot 2019-03-24 at 10 29 57 PM

- [x] Yes
- [ ] No

aggarwalpulkit596:
Fixes #1402 

Changes: Improvise build time

Screenshots for the change:
Screenshot 2019-03-24 at 10 24 12 PM


haroldadmin:
Enabling the Gradle daemon is not viable for everyone. Especially if you're running on a machine with 8 gigs of RAM or less. This is not the way to improve the build speed. I have two machines, one mac and one linux, both having 8 GB of RAM. The Gradle daemon makes them nearly unusable when enabled, even on multi module projects which have parallelized builds. I strongly recommend against this decision.

A better method would be to modularize the project.

aggarwalpulkit596:
I am currently using a mac with 8 gigs of ram and it works just fine and it's not just that there is also configure on demand and increasing the heap size @iamareebjamal what are your thoughts on this

haroldadmin:
What about the people with 4 Gigs of RAM? Or people who do other things on their system while also running Android Studio? IMO Gradle daemon is a strictly local setting. Everyone should configure it for their own machines, instead of forcing everyone to use the same settings.

iamareebjamal:
Yes, this does not improve the CI build time and Android Studio automatically utilizes daemon on supported machines

Issue #1436, opened 2019-03-26
Back navigation in ActionBar doesn't work when "difficulty fetching event" displayed


anhanh11001:
**Describe the bug**
Back navigation in header doesn't work when "difficulty fetching event" displayed

**Screenshots**
<img src="https://i.ibb.co/tcDsNV1/ezgif-2-540e4538c96d.gif" width="300">

**Would you like to work on the issue?**
Yes

PR #1442, opened 2019-03-26
fix: back button in actionbar

anhanh11001:
Details:
Previously, when the error fetching message is displayed, the other menu items in the action bar are hided by setting setHasOptionMenu to false. This made the android.R.id.home menu item stop working.

So instead of setting the setHasOptionMenu to false, setting the menu item visibility to false solve problem

Fixes: #1436

Screenshots for the change:


app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -394,5 +393,9 @@ class EventDetailsFragment : Fragment() {
     private fun showEventErrorScreen(show: Boolean) {
         rootView.container.visibility = if (!show) View.VISIBLE else View.GONE
         rootView.eventErrorCard.visibility = if (show) View.VISIBLE else View.GONE
+        val menuItemSize = menuActionBar?.size() ?: 0
+        for (i in 0..(menuItemSize - 1)) {
+            menuActionBar?.getItem(i)?.isVisible = !show
+        }
```
iamareebjamal: And when we go back?

anhanh11001: Sorry but I don't really understand your question?

iamareebjamal: Does it reappear correctly when we go back?

anhanh11001: As far as I understand, it will reappear correctly because: 
1. Menu Items visibility will only change on function `showEventErrorScreen()`
2. When error fetching occurs, the error is observed and call that function `showEventErrorScreen()` to true -> menu items visibility is false -> disappear
3. When error fetching stop occurring, the event detail is fetched and the `event` variable in the ViewModel is observed and call that function `showEventErrorScreen()` to false -> menu items visibility is true -> reappear

You can see it in here from line 84 to 104
```
       eventViewModel.event
            .nonNull()
            .observe(this, Observer {
                loadEvent(it)
                eventShare = it
                title = eventShare.name

                if (eventShare.favorite) {
                    setFavoriteIcon(R.drawable.ic_baseline_favorite_white)
                }

                if (runOnce) {
                    loadSocialLinksFragment()
                    loadSimilarEventsFragment()
                }
                runOnce = false

                Timber.d("Fetched events of id %d", safeArgs.eventId)
                showEventErrorScreen(false)
                setHasOptionsMenu(true)
            })
```

Issue #1446, opened 2019-03-26
SocialLink is not fetched again when internet connection reappear


anhanh11001:
**Describe the feature you'd like**
In EventDetailsFragment, social links are not fetched and displayed again when the internet connection is ON, but instead, the progress bar is still visible

**Screenshots**
<img src="https://i.ibb.co/Mc6nV6b/ezgif-2-b407bbfb2f96.gif" width="300"/>

**Would you like to work on the issue?**
Yes

PR #1447, opened 2019-03-27
fix: Add no internet view on social link fragment

anhanh11001:
Details:
- Check for internet connection before fetching social links, if internet is not connected -> display no internet view else -> load social links
- Inside internet connection, there is an image button to reload when internet connected again.

Fixes: #1446

Screenshots for the change:


iamareebjamal:
These are micro enhancements. Good but not required for now.

Issue #1433, opened 2019-03-26
Dark icons on blue accent color


addiegupta:
**Describe the feature you'd like**

The Floating Action Buttons have dark icon on its blue accent and purple background which would look better if a lighter color like white is used 
as before. I would like to replace the black icons with white

**Screenshots**

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/22665789/54997004-c3f6fd80-4ff0-11e9-9260-dc6d433a975c.gif)

**Additional context**

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

PR #1448, opened 2019-03-27
fix: display white icons in accent color FABs

addiegupta:
Fixes: #1433

Changes: 

After setting up Material components in the project, the FAB icons started appearing black instead of white.
This commit turns them back into white be setting the default color to be used on secondary color views to be white.
However as this also includes white background FABs, black tint for such icons has also been applied
Additionally `android:focusable="true"` has been added to a FAB to remove warning

Screenshots for the change:

PR 1448


addiegupta:
@liveHarshit please review

app/src/main/res/values/styles.xml
```diff
@@ -6,6 +6,7 @@
         <item name="colorPrimary">@color/colorPrimary</item>
         <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
         <item name="colorAccent">@color/colorAccent</item>
+        <item name="colorOnSecondary">@android:color/white</item>
```
iamareebjamal: Just this should be enough

addiegupta: But then the white FABs used in event cards have white icons making the icons invisible

Issue #1474, opened 2019-03-30
App crash on going back from a similar event


addiegupta:
**Describe the bug**

The app crashes when a similar event is selected from `EventDetailsFragment` for an event and then the up button or the back button is pressed.

**To Reproduce**

The app should not crash

**Logs**

2019-03-30 15:50:59.520 6773-6824/com.eventyay.attendee E/AndroidRuntime: FATAL EXCEPTION: pool-1-thread-1
    Process: com.eventyay.attendee, PID: 6773
    java.lang.IndexOutOfBoundsException: Index: 27, Size: 0
        at java.util.ArrayList.get(ArrayList.java:437)
        at androidx.recyclerview.widget.AsyncListDiffer$1$1.areItemsTheSame(AsyncListDiffer.java:306)
        at androidx.recyclerview.widget.DiffUtil.diffPartial(DiffUtil.java:268)
        at androidx.recyclerview.widget.DiffUtil.calculateDiff(DiffUtil.java:145)
        at androidx.recyclerview.widget.DiffUtil.calculateDiff(DiffUtil.java:105)
        at androidx.recyclerview.widget.AsyncListDiffer$1.run(AsyncListDiffer.java:292)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)

**Screenshots**

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/22665789/55274907-0e83bd00-5304-11e9-9362-dd7d3e540d27.gif)

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Nokia 6.1 Plus|
|Android Version| Pie 9.0    |

**Additional context**

The app seems to crash for certain events and not all. If you are unable to reproduce this issue, try on some other event.

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

liveHarshit:
I think it is memory bounce issue.

addiegupta:
@liveHarshit I am getting this error

2019-03-31 11:54:39.269 8825-9677/com.eventyay.attendee E/AndroidRuntime: FATAL EXCEPTION: pool-1-thread-2
    Process: com.eventyay.attendee, PID: 8825
    java.lang.IndexOutOfBoundsException: Index: 24, Size: 12
[...]

PR #1489, opened 2019-03-31
fix: app crash on going back from similar fragment'

addiegupta:
Fixes: #1474

Changes:

- On selecting a similar event from the EventDetailsFragment, a newEventDetailsFragment is created for the selected event.
- However using Koin DI, the same adapter object was being used.
- This resulted in app crashes on going back from the new event.
- This commit fixes the issue by instantiating the SimilarEventsListAdapter using its constructor and not KOIN DI

Additional Changes

- Removed shuffling of list of Similar events so that the fragment returns to the original order on pressing back
- Corrected a typo in the word 'similar' in `SimilarEventsViewModel`

Screenshots for the change:

prsimilarfragmentscrash


addiegupta:
Nothing is breaking and I think this should be good for this release

Issue #1491, opened 2019-03-31
No internet view on TicketsFragment


anhanh11001:
**Describe the bug**
Add no internet view on TicketsFragment

**Screenshots**
<img src="https://i.ibb.co/58bpRcf/55608075-1213485758824214-3616519324027584512-n.png" width="300">
**Would you like to work on the issue?**
Yes

PR #1492, opened 2019-03-31
feat: Add no internet view on TicketsFragment

anhanh11001:
Detail:
Add no internet card on Ticket Fragment and show it when there is no internet and no tickets retained in view model

Fixes: #1491

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/ticket/TicketsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/ticket/TicketsFragment.kt
@@ -16,6 +16,8 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.navigation.Navigation.findNavController
 import androidx.navigation.fragment.navArgs
 import com.google.android.material.snackbar.Snackbar
+import kotlinx.android.synthetic.main.content_no_internet.view.retry
+import kotlinx.android.synthetic.main.content_no_internet.view.noInternetCard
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/ticket/TicketsFragment.kt
```diff
@@ -194,4 +200,22 @@ class TicketsFragment : Fragment() {
         val alert = builder.create()
         alert.show()
     }
+
+    private fun loadTickets() {
+        if (!isNetworkConnected(context) && ticketsViewModel.tickets.value.isNullOrEmpty())
+            showNoInternetScreen(true)
```
liveHarshit: Why you are showing no internet screen if tickets value is null or empty?

anhanh11001: Because if the tickets information is loaded and already retained in the ViewModel then I think we can just show that information, no need for no internet card

iamareebjamal: @anhanh11001 This is wrong. It should be shown from the ViewModel. I am merging this, but create an issue to move all network logic to VM. Use BroadcastReceiver to make it reactive

anhanh11001: I will do it

Issue #1499, opened 2019-04-01
Event Title text getting cut off on the right edge


addiegupta:
**Describe the bug**

The text of the event title in EventDetailsFragment is getting cut off on the right edge

**To Reproduce**

The text should not get cut off

**Screenshots**

<img src = "https://user-images.githubusercontent.com/22665789/55313824-5d596000-5486-11e9-81af-88d859e24b39.jpeg" width = '320' height = '676'/>

<img src = "https://user-images.githubusercontent.com/22665789/55313828-5fbbba00-5486-11e9-98e8-0cb7cb2a2ef7.jpeg" width = '320' height = '676'/>

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | Nokia 6.1 Plus|
|Android Version| Pie 9.0 |

**Additional context**

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

PR #1500, opened 2019-04-01
fix: event title being cut off in EventDetailsFragment

addiegupta:
Fixes: #1499

Changes:
- Add right margins and change other attributes according to constraint layout to correct the issue of text being cut off

Additional changes:
- Refactored a few layout view identifiers to remove ambiguity

Screenshots for the change:


addiegupta:
@liveHarshit please review

Issue #1139, opened 2019-02-20
Bottom Navigation issues


iamareebjamal:
- [ ] Each time I go to events section, it is reloading
- [ ] Each time I go to Tickets section, it is reloading

This should not happen

liveHarshit:
Issue #1078 for events section.

angmas1:
Can I work on this?

anhanh11001:
@iamareebjamal @liveHarshit I have investigated this issue over the last 2 days. For my own knowledge now, it's clueless for me to fix this issue.

So basically the problem I found is that in the MainActivity, we set up the bottom navigation with navController:
`setupWithNavController(navigation, navController)`
From what I understand, everytime we click on a bottom navigation's icon, the navController will use the method navigate() to change the fragment in the container, and then clear(pop) all the fragment in the backstack. So if we use navigation with navController here, I don't find a way to keep all the old fragments.

I have tried to research on Singleton pattern to make each fragment a single instance, but I didn't find anything helpful. I also tried to save old fragments by using supportFragmentManager to save the fragment with tag but it made the code really complex and it would take a lot of time to refractor.

Please give some advices or correct me if something is not right

liveHarshit:
> So basically the problem I found is that in the MainActivity, we set up the bottom navigation with navController:
> `setupWithNavController(navigation, navController)`
> From what I understand, everytime we click on a bottom navigation's icon, the navController will use the method navigate() to change the fragment in the container, and then clear(pop) all the fragment in the backstack. So if we use navigation with navController here, I don't find a way to keep all the old fragment

Please read comments in issue #1078. It is already discussed.

liveHarshit:
> I also tried to save old fragments by using supportFragmentManager to save the fragment

It was used earlier, but now we want to do it with navigation architecture component. 
You can see the method in utils for check and load fragment. It is still there.

anhanh11001:
is anyone working on these issues? Can I work on this?

sdew021:
hey @iamareebjamal i see that this issue is still open, if no one has picked it, can i take up this issue and start working on it

iamareebjamal:
yes if you can still reproduce it

YugN17:
i want to work on this
can you assign me

YugN17:
@mariobehling i want to work on this can you assign  me

iamareebjamal:
We don't assign issues. Please send a PR

PR #1501, opened 2019-04-01
fix: Retain fragments on bottom navigation on nav controller

anhanh11001:
Details:
An Advanced Navigation workaround from Google is used to retain all of the fragments. Each of the item in bottom navigation is retained as navigation graph (5 navigation graphs - Events, Search, Likes, Tickets, Profile).

Some of the changes are made for proper navigation between fragments from different navigation graph (Example: booking order finished in EventsFragment - Events navigation graph to seeing ticket in OrderUnderUserFragment - Tickets navigation graph)

In LoginFragment, if (loginViewModel.isLoggedIn()) popBackStack() is moved to onActivityCreated as popBackStack() function require rootView to be created.

In SearchFragment, queryListener is moved to onCreateView is because onPrepareOptionMenu is not called every time we navigate to SearchFragment by bottom navigation, only the first time.

In SearchLocationFragment, fragmentManager is replaced by childFragmentManager because PlaceAutocompleteFragment is a fragment used inside a fragment, so using fragmentManager would produce "No view found for id 0x7f090065 for fragment PlaceAutocompleteFragment

Fixes: #1139 

Screenshots for the change:
Here is the [youtube link](https://www.youtube.com/watch?v=bGX_lnJYPvw) for preview as it is a large video, I cannot make GIFs


anhanh11001:
@liveHarshit @nikit19  please review!

anhanh11001:
@liveHarshit well then it's hard, in order to retain the data, at least the ViewModel must be retained, but in order for ViewModel to be retained then the fragments must survive when we navigate to another fragment by bottom navigation. However, I think current way to set up navigation with navigation component doesn't support retaining old fragments when navigating to another with bottomNavigationView and NavController. That's why this PR is large as I change the way navigation works with an extension function from Google sample

mariobehling:
OK, let's merge this. Please resolve conflicts.

Issue #1505, opened 2019-04-01
Add Menu Option to navigate to External Event Url from EventDetailsFragment


addiegupta:
**Describe the feature you'd like**

I would like to add a menu option to navigate to external event url so that the additional details about event can also be viewed.

**Screenshots**

**Additional context**

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

PR #1508, opened 2019-04-02
feat: add external event url menu option in EventDetailsFragment

addiegupta:
Fixes: #1505

Changes:

- Added a menu option to EventDetailsFragment to navigate to the external event url for the event
- If the external URL is null, then the menu option is disabled

Screenshots for the change:

prfor1505

Additional details:
- The option has been titled 'Open External Event URL' to maintain consistency with the eventyay frontend e.g. check the option on the right side at https://eventyay.com/e/a1fcaa21/coc/

- Instead of removing the menu item if the URL is null, I have chosen to disable the option because removing the item requires calling `invalidateOptionsMenu()` which will force the menu options to be prepared once again. This resets the favorite icon status also hence the favorite icon will have to be set again. Hence to avoid this complexity I have simply disabled it. If suggested, I can use this approach to remove the option altogether when the URL is null

EDIT: Updated GIF after now hiding option when URL is null
Will squash commits once this gets approved

prfor1505new

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -54,6 +54,7 @@ import org.fossasia.openevent.general.event.EventUtils.loadMapUrl
 import org.fossasia.openevent.general.event.topic.SimilarEventsFragment
 import org.fossasia.openevent.general.social.SocialLinksFragment
 import org.fossasia.openevent.general.ticket.TicketsFragmentArgs
+import org.fossasia.openevent.general.utils.Utils
 import org.fossasia.openevent.general.utils.Utils.getAnimSlide
 import org.fossasia.openevent.general.utils.Utils.requireDrawable
 import org.fossasia.openevent.general.utils.extensions.nonNull
@@ -91,9 +92,8 @@ class EventDetailsFragment : Fragment() {
                 eventShare = it
                 title = eventShare.name
 
-                if (eventShare.favorite) {
-                    setFavoriteIcon(R.drawable.ic_baseline_favorite_white)
-                }
+                // Update favorite icon and external event url menu option
+                activity?.invalidateOptionsMenu()
 
                 if (runOnce) {
                     loadSocialLinksFragment()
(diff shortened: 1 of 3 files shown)


liveHarshit:
Hide the option if url not available.

liveHarshit:
> * can use this approach to remove the option altogether when the URL is null

IMO, hiding it is a better option.

addiegupta:
@liveHarshit changes made

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -352,6 +351,18 @@ class EventDetailsFragment : Fragment() {
...
+        if (::eventShare.isInitialized) {
+
+            if (eventShare.externalEventUrl == null) menu.findItem(R.id.open_external_event_url).isVisible = false
+
+            setFavoriteIconFilled(eventShare.favorite)
+        }
+
+        super.onPrepareOptionsMenu(menu)
+    }
+
```
liveHarshit: Remove extra lines.

addiegupta:
@liveHarshit removed extra lines and squashed commits

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -352,6 +351,16 @@ class EventDetailsFragment : Fragment() {
         menuActionBar = menu
     }
 
+    override fun onPrepareOptionsMenu(menu: Menu) {
+        if (::eventShare.isInitialized) {
+            if (eventShare.externalEventUrl == null) {
```
nikit19: These two if statements can be combined

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -352,6 +351,16 @@ class EventDetailsFragment : Fragment() {
...
+                menu.findItem(R.id.open_external_event_url).isVisible = false
+            }
+            setFavoriteIconFilled(eventShare.favorite)
```
addiegupta: @nikit19 this statement is also to be executed only if `eventShare` has been initialised

Issue #1142, opened 2019-02-21
App crashes on API 19 Kitkat


angmas1:
**Describe the bug**
App crashes on kitkat API 19.

**To Reproduce**
1. Open app on kitkat.
2. Crashes.

**Expected behavior**
App shouldn't crash.

**Logs**
02-21 17:05:47.110 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'androidx.core.view.ViewCompat$2', referenced from method androidx.core.view.ViewCompat.addOnUnhandledKeyEventListener
02-21 17:05:47.110 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'android.view.WindowInsets', referenced from method androidx.core.view.ViewCompat.dispatchApplyWindowInsets
02-21 17:05:47.120 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'android.view.WindowInsets', referenced from method androidx.core.view.ViewCompat.onApplyWindowInsets
02-21 17:05:47.120 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'android.view.View$OnUnhandledKeyEventListener', referenced from method androidx.core.view.ViewCompat.removeOnUnhandledKeyEventListener
02-21 17:05:47.120 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'androidx.core.view.ViewCompat$1', referenced from method androidx.core.view.ViewCompat.setOnApplyWindowInsetsListener
02-21 17:05:47.130 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'android.graphics.drawable.RippleDrawable', referenced from method androidx.appcompat.widget.AppCompatImageHelper.hasOverlappingRendering
02-21 17:05:47.140 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'android.view.textclassifier.TextClassificationManager', referenced from method androidx.appcompat.widget.AppCompatTextClassifierHelper.getTextClassifier
02-21 17:05:47.280 3875-3875/com.eventyay.attendee E/dalvikvm: Could not find class 'java.beans.Transient', referenced from method com.fasterxml.jackson.databind.ext.Java7SupportImpl.<init>
[...]

PR #1536, opened 2019-04-04
chore: Raise minSdk to Lollipop

addiegupta:
Fixes: #1142

Changes:
- Updated minSdkVersion to 21 from 16 as OkHttp no longer supports API levels less than 21
- Removed unnecessary checks for sdk version >=21
- Removed 'left' and 'right' and replaced them with 'start' and 'end' in layout files as left/right are not needed in API 21 and start/end are recommended due to their RTL support

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -15,7 +15,7 @@ android {
     compileSdkVersion 28
     defaultConfig {
         applicationId "com.eventyay.attendee"
-        minSdkVersion 16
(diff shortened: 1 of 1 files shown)


iamareebjamal:
@addiegupta Revert all changes except minSdk

We want to be able to support KitKat if need be

addiegupta:
Closed by mistake

Issue #1544, opened 2019-04-05
Chip (Weekend, next month, today, tomorrow) object is still clickable when events are being fetched


anhanh11001:
**Describe the bug**
Chip objects in SearchFragment is still clickable when event is loading

**Screenshots**
<img src="https://i.ibb.co/p4WWyHn/ezgif-4-83a515575b98.gif" width="300">

**Would you like to work on the issue?**
Yes

PR #1549, opened 2019-04-05
fix: Make chip not clickable while fetching events

anhanh11001:
Detail:
- A clickable flag inside SearchViewModel is used to check whether chip should be clickable or not.
- Change id name of chip in XML file

Fixes: #1544

Screenshots for the change:
When loading, those chips card are not clickable any more


anhanh11001:
resolved

PR #1636, opened 2019-04-20
chore: Move flask filter logic from ViewModels to Services

### LINKED ISSUE #1344: chore: Move flask filter query logic from ViewModels to Services
**Describe the chore**
Some `ViewModel`s form the flask filter query themselves and then pass it on to their correspoding `Service`s. An example would be `EventsViewModel` forming the query in `loadLocationEvents` method and passing it to `EventService`.

The problem is that this moves the logic of retrieving data from `Services`(where it belongs) to `ViewModel`s. The `ViewModel`s should not be concerned with how the data is being retrieved. That is `Service`s role.

So, the flask filter query should be formed in the `Service`s and not in the `ViewModel`s.

- [X] Yes
- [ ] No

aggarwalpulkit596:
Fixes #1344

Changes: Earlier, ViewModels were building flask filter queries themselves and passing them to Services.

This was not good because this leaked the data retrieval logic into the ViewModels. Only the Services should be concerned with data retrieval and they should abstract out that logic.

This change moves the logic into Services where it belongs.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeService.kt
@@ -27,12 +27,13 @@ class AttendeeService(
         return attendeeApi.deleteAttendee(id)
     }
 
-    fun getCustomFormsForAttendees(id: Long, filter: String): Single<List<CustomForm>> {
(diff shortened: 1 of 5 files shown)


aggarwalpulkit596:
@iamareebjamal resolved conflicts

Issue #1655, opened 2019-04-23
Chip in FavoriteFragment is checkable and search results are not fetched


anhanh11001:
**Describe the bug**
In FavoriteFragment, chips are checkable and currently, when clicking on a chip, results are not fetched

**Screenshots**
<img src="https://i.imgur.com/1j5AcL8.gif">

**Would you like to work on the issue?**
Yes

PR #1659, opened 2019-04-24
fix: Make chip not checkable and search possible

anhanh11001:
Detail:
- In FavoriteFragment, make chips not clickable
- Make search possible by adding search type

Fixes: #1655

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
@@ -150,7 +150,10 @@ class FavoriteFragment : Fragment() {
     }
 
     private fun openSearchResult(time: String) {
-        SearchResultsFragmentArgs(query = "", location = Preference().getString(SAVED_LOCATION).toString(), date = time)
+        SearchResultsFragmentArgs(query = "",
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/favorite/FavoriteFragment.kt
```diff
@@ -153,6 +153,7 @@ class FavoriteFragment : Fragment() {
     private fun openSearchResult(time: String) {
         SearchResultsFragmentArgs.Builder()
             .setQuery("")
+            .setType("Anything")
```
aggarwalpulkit596: Why hardcoding the type there's no point of chips if you're doing this you should make them as they were supposed to work instead of making a dirty fix

anhanh11001: @aggarwalpulkit596  I don't really get your point... So the problem is simply because the type in SearchResultFragment is set to null when being navigated from this, so results are not fetched. If I don't put the default type ("Anything") here, then sure I can put it in the SearchResultFragment, and that's pretty much the same

anhanh11001:
@liveHarshit please review

Issue #1656, opened 2019-04-23
EventDetailFragment automatically move to a section without scrolling


anhanh11001:
**Describe the bug**
In EventDetailFragment, sometimes it automatically moved to a below section before user trying to scroll or do something

**Screenshots**
<img src="https://i.imgur.com/IZIp2Ob.gif">

**Would you like to work on the issue?**
Yes

aggarwalpulkit596:
Just add android:descendantFocusability="blocksDescendants in the organiser container

PR #1662, opened 2019-04-24
fix: Block descendant focusability on sections in event detail fragment

anhanh11001:
Block descendant focusability on sections in event detail fragment

Fixes: #1656


iamareebjamal:
What was the issue? Can you explain?

aggarwalpulkit596:
@iamareebjamal eventDetailsFragment was scrolling down to organizer Container Automatically

PR #1681, opened 2019-04-28
feat: Add Pagination

### LINKED ISSUE #84: Implement pagination in the app
Working on it

aggarwalpulkit596:
Fixes #84 

Changes: Add Pagination for Searching Location Events

Screenshots for the change:
Infinite Scrolling
videotogif_2019 05 02_19 17 29

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,12 +36,12 @@ android {
             shrinkResources true
             minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
-            buildConfigField "String", "DEFAULT_BASE_URL", '"https://api.eventyay.com/v1/"'
             buildConfigField "String", "MAPBOX_KEY", '"'+MAPBOX_KEY+'"'
+            buildConfigField "String", "DEFAULT_BASE_URL", '"https://open-event-api-dev.herokuapp.com/v1/"'
             buildConfigField "String", "STRIPE_API_KEY", '"'+STRIPE_API_TOKEN+'"'
         }
         debug {
-            buildConfigField "String", "DEFAULT_BASE_URL", '"https://open-event-api-dev.herokuapp.com/v1/"'
+            buildConfigField "String", "DEFAULT_BASE_URL", '"https://api.eventyay.com/v1/"'
             buildConfigField "String", "MAPBOX_KEY", '"'+MAPBOX_KEY+'"'
             buildConfigField "String", "STRIPE_API_KEY", '"'+STRIPE_API_TOKEN+'"'
         }
@@ -156,6 +156,10 @@ dependencies {
     implementation 'io.reactivex.rxjava2:rxkotlin:2.3.0'
     implementation 'io.reactivex.rxjava2:rxjava:2.2.8'
     implementation 'com.squareup.retrofit2:adapter-rxjava2:2.4.0'
+
+    //Paging
+    implementation "android.arch.paging:rxjava2:1.0.1"
+    implementation "android.arch.paging:runtime:1.0.1"
 
     // Picasso
     implementation 'com.squareup.picasso:picasso:2.71828'
--- a/app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.general.di
 
+import androidx.paging.PagedList
 import androidx.room.Room
 import com.facebook.stetho.okhttp3.StethoInterceptor
 import com.fasterxml.jackson.databind.DeserializationFeature
@@ -201,7 +202,7 @@ val apiModule = module {
 
 val viewModelModule = module {
     viewModel { LoginViewModel(get(), get(), get()) }
-    viewModel { EventsViewModel(get(), get(), get(), get()) }
+    viewModel { EventsViewModel(get(), get(), get(), get(), get()) }
     viewModel { ProfileViewModel(get(), get()) }
     viewModel { SignUpViewModel(get(), get(), get()) }
     viewModel { EventDetailsViewModel(get(), get(), get(), get(), get(), get(), get(), get()) }
@@ -235,6 +236,16 @@ val networkModule = module {
         val objectMapper = jacksonObjectMapper()
         objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
         objectMapper
+    }
+
+    single {
+        PagedList
+            .Config
+            .Builder()
+            .setPageSize(10)
+            .setInitialLoadSizeHint(10)
+            .setEnablePlaceholders(false)
+            .build()
     }
 
     single { RequestAuthenticator(get()) as Interceptor }
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
@@ -3,10 +3,10 @@ package org.fossasia.openevent.general.event
 import io.reactivex.Completable
 import io.reactivex.Flowable
 import io.reactivex.Single
-import org.fossasia.openevent.general.event.feedback.Feedback
-import org.fossasia.openevent.general.event.feedback.FeedbackApi
 import org.fossasia.openevent.general.event.faq.EventFAQ
 import org.fossasia.openevent.general.event.faq.EventFAQApi
+import org.fossasia.openevent.general.event.feedback.Feedback
+import org.fossasia.openevent.general.event.feedback.FeedbackApi
 import org.fossasia.openevent.general.event.location.EventLocation
 import org.fossasia.openevent.general.event.location.EventLocationApi
 import org.fossasia.openevent.general.event.topic.EventTopic
@@ -17,6 +17,8 @@ import org.fossasia.openevent.general.event.types.EventTypesApi
 import org.fossasia.openevent.general.speakercall.SpeakersCall
 import org.fossasia.openevent.general.speakercall.SpeakersCallDao
 import java.util.Date
+import kotlin.collections.ArrayList
+import kotlin.collections.set
 
 class EventService(
     private val eventApi: EventApi,
@@ -109,6 +111,15 @@ class EventService(
                 eventDao.getEventWithIds(eventIds)
             }
     }
+
+    fun getEventsByLocationPaged(locationName: String?, page: Int): Flowable<List<Event>> {
+        val query = "[{\"name\":\"location-name\",\"op\":\"ilike\",\"val\":\"%$locationName%\"}," +
+            "{\"name\":\"ends-at\",\"op\":\"ge\",\"val\":\"%${EventUtils.getTimeInISO8601(Date())}%\"}]"
+        return eventApi.searchEventsPaged("name", query, page).flatMapPublisher { apiList ->
+            updateFavorites(apiList)
+        }
+    }
+
 
     fun getEvent(id: Long): Flowable<Event> {
         return eventDao.getEvent(id)
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsDataSource.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsDataSource.kt
@@ -1,0 +1,54 @@
+package org.fossasia.openevent.general.event
+
+import androidx.lifecycle.MutableLiveData
+import androidx.paging.PageKeyedDataSource
+import io.reactivex.disposables.CompositeDisposable
+import io.reactivex.rxkotlin.plusAssign
+import org.fossasia.openevent.general.utils.extensions.withDefaultSchedulers
+
+class EventsDataSource(
+    private val eventService: EventService,
+    private val compositeDisposable: CompositeDisposable,
+    private val query: String?,
+    private val mutableProgress: MutableLiveData<Boolean>
+
+) : PageKeyedDataSource<Int, Event>() {
+    override fun loadInitial(
+        params: LoadInitialParams<Int>,
+        callback: LoadInitialCallback<Int, Event>
+    ) {
+        createObservable(1, 2, callback, null)
+    }
+
+    override fun loadAfter(params: LoadParams<Int>, callback: LoadCallback<Int, Event>) {
+        val page = params.key
+        createObservable(page, page + 1, null, callback)
+    }
+
+    override fun loadBefore(params: LoadParams<Int>, callback: LoadCallback<Int, Event>) {
+        val page = params.key
+        createObservable(page, page - 1, null, callback)
+    }
+
+    private fun createObservable(
+        requestedPage: Int,
+        adjacentPage: Int,
+        initialCallback: LoadInitialCallback<Int, Event>?,
+        callback: LoadCallback<Int, Event>?
+    ) {
+        compositeDisposable +=
+            eventService.getEventsByLocationPaged(query, requestedPage)
+                .withDefaultSchedulers()
+                .subscribe(
+                    { response ->
+                        if (requestedPage == 1) {
+                            mutableProgress.value = false
+                        }
+                        initialCallback?.onResult(response, null, adjacentPage)
+                        callback?.onResult(response, adjacentPage)
+                    },
+                    { error ->
+                    }
+                )
+    }
+}
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsDataSourceFactory.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsDataSourceFactory.kt
@@ -1,0 +1,21 @@
+package org.fossasia.openevent.general.event
+
+import androidx.lifecycle.MutableLiveData
+import androidx.paging.DataSource
+import io.reactivex.disposables.CompositeDisposable
+
+class EventsDataSourceFactory(
+    private val compositeDisposable: CompositeDisposable,
+    private val eventService: EventService,
+    private val query: String?,
+    private val mutableProgress: MutableLiveData<Boolean>
+) : DataSource.Factory<Int, Event>() {
+    override fun create(): DataSource<Int, Event> {
+        return EventsDataSource(
+            eventService,
+            compositeDisposable,
+            query,
+            mutableProgress
+        )
+    }
+}
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
@@ -1,10 +1,16 @@
 package org.fossasia.openevent.general.event
 
 import androidx.lifecycle.LiveData
+import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import androidx.paging.PagedList
+import androidx.paging.RxPagedListBuilder
+import io.reactivex.Observable
+import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.rxkotlin.plusAssign
+import io.reactivex.schedulers.Schedulers
 import org.fossasia.openevent.general.R
 import org.fossasia.openevent.general.common.SingleLiveEvent
 import org.fossasia.openevent.general.connectivity.MutableConnectionLiveData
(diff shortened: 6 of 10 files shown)


aggarwalpulkit596:
@iamareebjamal  any suggestions how to pass search query into Data source Factory for pagedList api call

app/src/main/java/org/fossasia/openevent/general/event/EventsDataSourceFactory.kt
```diff
@@ -0,0 +1,20 @@
+package org.fossasia.openevent.general.event
+
+import androidx.lifecycle.MutableLiveData
+import androidx.paging.DataSource
+import io.reactivex.disposables.CompositeDisposable
+import org.fossasia.openevent.general.event.faq.EventsDataSource
+
+class EventsDataSourceFactory(private val compositeDisposable: CompositeDisposable,
+                              private val eventService: EventService,
+                              private val query:String?)
```
iamareebjamal: You are already doing it here

aggarwalpulkit596: I thought it can be some better way of doing this, i was not satisfied with the current implementation

iamareebjamal: Factory means you can take any argument to produce other type

aggarwalpulkit596:
@iamareebjamal @nikit19 please do an initial review after that I'll be moving forward to integrate for similar and search event API calls

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -18,32 +23,46 @@ class EventsViewModel(
...
+            )
+            mutableEventsPaged = RxPagedListBuilder(sourceFactory, config)
+                .setFetchScheduler(Schedulers.io())
+                .buildObservable()
+                .cache()
+
+            compositeDisposable += mutableEventsPaged
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
                 .doOnSubscribe {
```
aggarwalpulkit596: doFinally Block is not working

app/src/main/java/org/fossasia/openevent/general/event/EventsListAdapter.kt
```diff
@@ -57,7 +59,7 @@ class EventsListAdapter(
      * The function to call when the adapter has to be cleared of items
      */
     fun clear() {
-        this.submitList(emptyList())
+//        this.submitList()
```
aggarwalpulkit596: how to pass empty paged list I couldn't get it

app/build.gradle
```diff
@@ -36,11 +36,11 @@ android {
             shrinkResources true
             minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
-            buildConfigField "String", "DEFAULT_BASE_URL", '"https://api.eventyay.com/v1/"'
+            buildConfigField "String", "DEFAULT_BASE_URL", '"https://open-event-api-dev.herokuapp.com/v1/"'
```
aggarwalpulkit596: will be reverting this

aggarwalpulkit596:
@nikit19 any update on this

aggarwalpulkit596:
@iamareebjamal any update regarding this ?

iamareebjamal:
Take peer reviews

aggarwalpulkit596:
@nikit19 @liveHarshit please review

app/src/main/java/org/fossasia/openevent/general/event/EventsListAdapter.kt
liveHarshit: What is this `emptyList()` method returning earlier?

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
liveHarshit: Cannot get it, why it is not working!

app/src/main/java/org/fossasia/openevent/general/event/topic/SimilarEventsFragment.kt
```diff
@@ -159,8 +159,8 @@ class SimilarEventsFragment : Fragment() {
 
         handleVisibility(similarEvents)
         Timber.d("Fetched Similar events of size %s", similarEvents.size)
-        if (similarEventsListAdapter.currentList.size != similarEvents.size) similarEvents.shuffle()
-        similarEventsListAdapter.submitList(similarEvents)
+//        if (similarEventsListAdapter.currentList.size != similarEvents.size) similarEvents.shuffle()
+//        similarEventsListAdapter.submitList(similarEvents)
```
liveHarshit: Why?

aggarwalpulkit596: Will implement pagination here as well it uses same adapters event view holder

aggarwalpulkit596: @iamareebjamal can you tell why doFinally Block is not working here because of that shimmer layout do not stop everything else just works fine. is it because pagination requires continuous API calls to different pages?

app/src/main/java/org/fossasia/openevent/general/event/EventsListAdapter.kt
aggarwalpulkit596: empty list returns a list whereas the adapter required a pagedlist

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
aggarwalpulkit596: @iamareebjamal can you tell why doFinally Block is not working here because of that shimmer layout do not stop everything else just works fine. is it because pagination requires continuous API calls to different pages?

app/src/main/java/org/fossasia/openevent/general/event/EventsListAdapter.kt
aggarwalpulkit596: thinking to use     
` this.currentList?.clear() `
would that work @liveHarshit

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
aggarwalpulkit596: @iamareebjamal please a little bit of help here

iamareebjamal: Will try locally today

aggarwalpulkit596: @iamareebjamal any luck with this ?

iamareebjamal: Didn't get time. Hopefully tomorrow

aggarwalpulkit596: @iamareebjamal any update? 😅 i can try to find out someting myself if you're busy

iamareebjamal: Yes, please do. I'm not getting any time these days

aggarwalpulkit596: @iamareebjamal i tried with calling doFinally in the data source and it was working there,i might be right that it RxPagedList don't call doFinally as it subscribes again and again to fetch all the pages

aggarwalpulkit596: i was searching for some solution the only way i found is to send a mutableprogress object to data source and update that on doFinally

aggarwalpulkit596: I've pushed it but I don't think so it is the best possible method to implement it

iamareebjamal: Yeah, doFinally will never be called in case of Observable or Flowable because the stream is never completed

app/src/main/java/org/fossasia/openevent/general/event/EventsDataSource.kt
```diff
@@ -0,0 +1,58 @@
...
+        requestedPage: Int,
+        adjacentPage: Int,
+        initialCallback: LoadInitialCallback<Int, Event>?,
+        callback: LoadCallback<Int, Event>?
+    ) {
+        compositeDisposable +=
+            eventService.getEventsByLocationPaged(query, requestedPage)
+                .withDefaultSchedulers()
+                .doFinally {
+                    mutableProgress.value = false
```
aggarwalpulkit596: would this approach be okay? i couldn't find any other method to get this working @iamareebjamal

iamareebjamal: What is `getEventsByLocationPaged` returning?

aggarwalpulkit596: It is returning a Observable Paged List

iamareebjamal: You cannot use doFinally here, move this call inside subscribe

aggarwalpulkit596: Okay but inside the individual call of page finally is working

aggarwalpulkit596: Should i add a check using using pageNo if equal pageNo   == 1 then update the value because we want to hide the shimmer only once

iamareebjamal: Yes, that's reasonable

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -60,6 +74,12 @@ class EventsViewModel(
         } else {
             mutableProgress.value = false
         }
+
+        mutableProgress.observeForever {
```
aggarwalpulkit596: @iamareebjamal I can't observer in the ViewModel i guess. can i ?

iamareebjamal: No, use MediatorLiveData

aggarwalpulkit596: @iamareebjamal updated

aggarwalpulkit596: @iamareebjamal should I implement pagination for search events as well?

iamareebjamal: Build is failing. Yeah, implement that in other PR

aggarwalpulkit596: There are some unrelated changes will undo them and fix spotless. Rest all good?

app/src/main/java/org/fossasia/openevent/general/event/topic/SimilarEventsFragment.kt
aggarwalpulkit596: @iamareebjamal should i implement pagination here as well because this method won't no longer work as the adapter expects as pagedlist and we are passing a normal list here

aggarwalpulkit596: And there would be a heavy network usage if I implement pagination here as well I suggest we should create a seperate adapter for this and use this as a normal list

aggarwalpulkit596: @iamareebjamal updated

iamareebjamal:
@anhanh11001 @liveHarshit Please test

app/src/main/java/org/fossasia/openevent/general/event/EventsDataSource.kt
```diff
@@ -0,0 +1,58 @@
...
+        initialCallback: LoadInitialCallback<Int, Event>?,
+        callback: LoadCallback<Int, Event>?
+    ) {
+        compositeDisposable +=
+            eventService.getEventsByLocationPaged(query, requestedPage)
+                .withDefaultSchedulers()
+                .subscribe(
+                    { response ->
+                        if (requestedPage == 1) {
+                            mutableProgress.value = false
```
liveHarshit: Where is the progress set to true?

liveHarshit:
@iamareebjamal @aggarwalpulkit596 App is crashing while navigating to another fragment and press back button while loading - 
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.eventyay.attendee, PID: 6821
    java.lang.IllegalArgumentException: This source was already added with the different observer
        at androidx.lifecycle.MediatorLiveData.addSource(MediatorLiveData.java:89)
        at org.fossasia.openevent.general.event.EventsViewModel.loadLocationEvents(EventsViewModel.kt:78)
```

liveHarshit:
Also during loading events- 
image

aggarwalpulkit596:
@liveHarshit will check

aggarwalpulkit596:
> Also during loading events-
> image



> @iamareebjamal @aggarwalpulkit596 App is crashing while navigating to another fragment and press back button while loading -
> 
> ```
> E/AndroidRuntime: FATAL EXCEPTION: main
>     Process: com.eventyay.attendee, PID: 6821
>     java.lang.IllegalArgumentException: This source was already added with the different observer
>         at androidx.lifecycle.MediatorLiveData.addSource(MediatorLiveData.java:89)
>         at org.fossasia.openevent.general.event.EventsViewModel.loadLocationEvents(EventsViewModel.kt:78)
> ```

not able to replicate it @liveHarshit

liveHarshit:
> not able to reproduce it @liveHarshit

While loading goes to notification fragment and press back button. It's crashing in my device. Let me check again.

aggarwalpulkit596:
> > not able to reproduce it @liveHarshit
> 
> While loading goes to notification fragment and press back button. It's crashing in my device. Let me check again.
 Yup i got it

liveHarshit:
> Let me check again.

same

aggarwalpulkit596:
```
        eventsViewModel.progress
            .nonNull()
                rootView.swiperefresh.isRefreshing = it
            })
```
@liveHarshit i need to move this code to onCreate as the observer is not removed and ViewModel is not destroyed

aggarwalpulkit596:
Or i have to manually remove the observer in onDestroy or orPause method

liveHarshit:
@aggarwalpulkit596 Also check how swipe refresh is working. It is showing no events instead of shimmer during refreshing.

liveHarshit:
> i need to move this code to onCreate as the observer is not removed and ViewModel is not destroyed

In onCreate `viewLifecycleOwner` is null.

aggarwalpulkit596:
@liveHarshit will find another solution to it 👍

aggarwalpulkit596:
@liveHarshit @iamareebjamal please review there is only issue now

> Also during loading events-
> image

this always shows up because previously observer was only called upon a valid api request not with paged list it is called as soon api call is made so we need to check additional param whether showShimmer is false or not, just an observable on event list can't work

liveHarshit:
Please fix Travis.

aggarwalpulkit596:
Probably a spotless fix will fix it but please first check for any crashes 😅

liveHarshit:
@aggarwalpulkit596 It is not showing upcoming events.

liveHarshit:
> check how swipe refresh is working. It is showing no events instead of shimmer during refreshing.

also

aggarwalpulkit596:
> @aggarwalpulkit596 It is not showing upcoming events.

Oh the logic must have been changed i forgot will change the end point

aggarwalpulkit596:
> > check how swipe refresh is working. It is showing no events instead of shimmer during refreshing.
> 
> also

Forgot that will fix it rest all good ?

liveHarshit:
I don't see Infinite Scrolling in my device.

aggarwalpulkit596:
Did you try with changing the base url ther aren't many events in the staging server try with prod server

liveHarshit:
> Did you try with changing the base url ther aren't many events in the staging server try with prod server

Yes I have tried with Singapore and Delhi location with the release API. For what location you are testing?

aggarwalpulkit596:
Wait let me share a gif with you

aggarwalpulkit596:
videotogif_2019 05 19_19 35 24
@liveHarshit

app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -65,28 +88,14 @@ class EventsViewModel(
     fun isConnected(): Boolean = mutableConnectionLiveData.value ?: false
 
     fun clearEvents() {
-        mutableEvents.value = null
+
+       // mutableEventsPaged
```
aggarwalpulkit596: @iamareebjamal couldn't find anything to clear this out please help out here

app/src/main/java/org/fossasia/openevent/general/event/EventService.kt
```diff
@@ -99,6 +99,18 @@ class EventService(
         }
     }
 
+    fun getEventsByLocationPaged(locationName: String?, page: Int): Single<List<Event>> {
+        val query = "[{\"name\":\"location-name\",\"op\":\"ilike\",\"val\":\"%$locationName%\"}," +
```
aggarwalpulkit596: @liveHarshit updated query for upcoming events

aggarwalpulkit596:
@liveHarshit rest all good ?

liveHarshit:
> @liveHarshit rest all good ?

What is the minimum number of events required for infinite scrolling?

liveHarshit:
Because after filtering events, I can list max 5 number of events and infinite scrolling is not working. 
Also showing no events card before loading events in rv.

aggarwalpulkit596:
> > @liveHarshit rest all good ?
> 
> What is the minimum number of events required for infinite scrolling?

It is 6 but can be changed in config object and it is showing because of

> @liveHarshit @iamareebjamal please review there is only issue now
> 
> > Also during loading events-
> > image
> 
> this always shows up because previously observer was only called upon a valid api request not with paged list it is called as soon api call is made so we need to check additional param whether showShimmer is false or not, just an observable on event list can't work

app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
```diff
@@ -232,6 +233,16 @@ val networkModule = module {
         objectMapper
     }
 
+    single {
+        PagedList
+            .Config
+            .Builder()
+            .setPageSize(10)
```
aggarwalpulkit596: @liveHarshit it is 10

liveHarshit: Okay. Then there are three bugs to solve that I've already told you.

aggarwalpulkit596: I already @iamareebjamal for a little bit of assistance as i am unable to find anything for @liveHarshit you can also suggest something if you find anything 😅

liveHarshit: Firstly you need to modify the logic for showing no events card and shimmer effect. Both are in conflicts right now.

liveHarshit: Also how the swipe refresh is connected with both.

aggarwalpulkit596: In a normal api call live data observer trigger when api response is made but with paged list it is triggered with a zero size list and so current implement fails and to overcome this we need and additional param that whether the intial request is completer or not

liveHarshit: > triggered with a zero size list

Oh, got it. So here we cannot compare the size of the list, but 

> Also how the swipe refresh is connected with both.

In that case, can we check if there are no events?

aggarwalpulkit596: Can only be checked after the first request is completed that is progress live data is set to false

liveHarshit: Then show shimmer until the request will complete.

aggarwalpulkit596: Also for swipe to refresh we need to clear the mutableEventsPaged but that is observable before it was a live data so its value can be set to null and I can't find anything to set it source factory to null or zero-sized list

aggarwalpulkit596: > Then show shimmer until the request will complete.

but that's the point I need to check 2 separate live data values and then work accordingly will find some workaround for that definitely but couldn't find anything for observable disposable.clear() would work I guess then?

aggarwalpulkit596:
@liveHarshit updated,everything is working fine now
Couldn't check though because test events have been deleted i guess

liveHarshit:
> @liveHarshit updated,everything is working fine now
> Couldn't check though because test events have been deleted i guess

Please test it by removing the filter and change the debug API to release.

aggarwalpulkit596:
upcoming event filter? Cool

anhanh11001:
Any updates on this? If not then I'm taking this up as it is an issue listed in GSoC Milestones. @aggarwalpulkit596

aggarwalpulkit596:
Would  wind up by end of the week

anhanh11001:
@aggarwalpulkit596 Thanks

Issue #1691, opened 2019-04-29
Add Session Fragment


anhanh11001:
**Describe the feature you'd like**
Add fragment to show detailed information of the event's session

**Would you like to work on the issue?**
Yes

Issue #1646, opened 2019-04-23
Add Session Section


aggarwalpulkit596:
**Describe the feature you'd like**
Show a list of all the sessions organized by the user along with track speaker and other information regarding the track

**Would you like to work on the issue?**
- [x] Yes

angmas1:
This a part of GSOC tasks, so shouldn't it be done by the person who gets selected @iamareebjamal

iamareebjamal:
All tasks written in GSoC proposal cannot be all implemented by a single person and there is a huge overlap in ideas.

PR #1693, opened 2019-04-30
feat: Add session fragment

anhanh11001:
Detail:
- Add session fragment to show all details of a session.
- Fix network call on session + adding sessions to database.

Fixes: #1691

Screenshots for the change:
ezgif-4-d82b10d96f83


aggarwalpulkit596:
@anhanh11001 i already completed this PR and was waiting for #1646 to get merged

app/src/main/java/org/fossasia/openevent/general/sessions/SessionWithEvent.kt
```diff
@@ -0,0 +1,27 @@
...
+    foreignKeys = [
+        ForeignKey(entity = Event::class,
+            parentColumns = ["id"],
+            childColumns = ["event_id"]),
+        ForeignKey(entity = Session::class,
+            parentColumns = ["id"],
+            childColumns = ["session_id"])
+    ]
+)
+data class SessionWithEvent(
```
aggarwalpulkit596: Why this class no same session can be in two different events  as it is 1:n relationship and not m:n relationship there's no point having this table previously i created this table for speaker because speakers can exist this way but sessions can never exist in this way

app/src/main/java/org/fossasia/openevent/general/sessions/sessiontype/SessionType.kt
```diff
@@ -1,17 +1,27 @@
...
     @Id(LongIdHandler::class)
+    @PrimaryKey
     val id: Long,
     val name: String,
     val length: String?,
-    val deletedAt: String?
+    val deletedAt: String?,
+    @ColumnInfo
+    @Relationship("session")
+    var session: Session? = null
```
aggarwalpulkit596: Remove this relationship also as this relationship will never be used and will create a circular dependency nothing else

app/src/main/java/org/fossasia/openevent/general/sessions/SessionWithEvent.kt
anhanh11001: Thanks for the review. My initial intention was to cache the sessions so that users don't need to fetch sessions multiple time, not to blindly copy speaker section. But I have rethought and I think it might be better to cache them inside the View Model. Let's me know if you have a better suggestion

aggarwalpulkit596: the point is about caching them the point is of the unnecessary table of SessionWithEvent sessions are always unique can't be mapped to two different events

anhanh11001: Yes I get your point, I'm fixing it

aggarwalpulkit596: And just for reference you don't have to write this query for this table you can simply use relation annotations provided by room and it does all the work by itself i use this query because it's just makes more sense

app/src/main/java/org/fossasia/openevent/general/sessions/Session.kt
```diff
@@ -1,16 +1,25 @@
...
 import com.github.jasminb.jsonapi.LongIdHandler
 import com.github.jasminb.jsonapi.annotations.Id
 import com.github.jasminb.jsonapi.annotations.Relationship
 import com.github.jasminb.jsonapi.annotations.Type
+import org.fossasia.openevent.general.sessions.microlocation.MicroLocation
+import org.fossasia.openevent.general.sessions.sessiontype.SessionType
 
 @Type("session")
 @JsonNaming(PropertyNamingStrategy.KebabCaseStrategy::class)
+@Entity
```
nikit19: Database schema must have changed? I guess you didn't commit that?

anhanh11001: I think you mean an autogenerated json file in schema folder. Should I also commit that?

anhanh11001: I'm uploading it now.

aggarwalpulkit596:
@anhanh11001 build is failing

anhanh11001:
@aggarwalpulkit596 I fixed the view. Please review

anhanh11001:
@iamareebjamal please review

anhanh11001:
@liveHarshit @nikit19 please review again, this PR has been here for a while

anhanh11001:
@iamareebjamal please review

Issue #1706, opened 2019-05-02
Wrong QR code on individual ticket holder


anhanh11001:
**Describe the bug**
Currently, different ticket holders with the same order are having the same QR Codes for their ticket. The cause of this issue is that QR code is currently made by order identifier, while in the server, the QR code is currently made by order identifier + ticket holder id.

**Screenshots**
<img src="https://i.ibb.co/jyB2FD4/58822222-374397796510437-5585571751170932736-n.jpg" width="300">           <img src="https://i.ibb.co/m8jwxp4/59342905-574158816412113-5890533034985259008-n.jpg" width="300">

**Would you like to work on the issue?**
Yes

anhanh11001:
@iamareebjamal I don't know how to fix this because, in the server, the QR code is made by qrcode library from python, while in the android, it is made by zxing library. I'm not sure if both libraries have the same method to make the QR image but it seems they don't. And the problem is both of libraries are not cross-platform. Please give your thoughts?

iamareebjamal:
> QR code is currently made by order identifier, while in the server, the QR code is currently made by order identifier + ticket holder id

Then use order identifier + ticket holder id

anhanh11001:
My point is that as I use the same string + setting the same QR_Version and error_correction, the library in the server creates a different QR Image than the library in the Android app.

iamareebjamal:
Having different QR code is not a problem unless they resolve to different strings. When scanned, they should generate same resultant code

anhanh11001:
@iamareebjamal I get it, I'll start testing now

PR #1711, opened 2019-05-04
fix: Change the QR value based on server

anhanh11001:
Change the QR code based on server.
I have tested so that even though the QR image mobile app is not the same as the QR image in the pdf file, the value given after scanning is still the same.

Fixes: #1706


iamareebjamal:
Great, can you check if it works in the Orga App? If it does not, can you create an issue there?

anhanh11001:
@iamareebjamal Yes i'll do it now

iamareebjamal:
@anhanh11001 Status?

Issue #1215, opened 2019-03-03
Enlarge QR code and increase brightness on clicking


ShridharGoel:
**Describe the feature you'd like**
Feature to enlarge QR code and increase device's brightness when the QR code in order details is clicked.

**Would you like to work on the issue?**
- [x] Yes
- [ ] No

PR #1727, opened 2019-05-08
feat: Add enlarged QR Image + Increase brightness

anhanh11001:
Detail:
- Set up click listener to enlarge QR Image
- Increate brightness when clicking on QR Image.

Fixes: #1215

Screenshots for the change:
Somehow my screen recorder doesn't record the change in brightness. But it increases the brightness to maximum when dialog appear and return to current brightness when dismissing the dialog
ezgif-4-ef9e53ad591c

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/order/OrderDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/order/OrderDetailsFragment.kt
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.general.order
 
+import android.app.AlertDialog
 import android.content.Intent
 import android.graphics.Bitmap
 import android.graphics.Canvas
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/fossasia/openevent/general/order/OrderDetailsFragment.kt
```diff
@@ -24,6 +25,7 @@ import kotlinx.android.synthetic.main.fragment_order_details.view.orderDetailCoo
 import kotlinx.android.synthetic.main.fragment_order_details.view.orderDetailsRecycler
 import kotlinx.android.synthetic.main.fragment_order_details.view.progressBar
 import kotlinx.android.synthetic.main.item_card_order_details.view.orderDetailCardView
+import kotlinx.android.synthetic.main.item_enlarged_qr.view.*
```
ShridharGoel: Please avoid wildcard imports.

ShridharGoel:
Shouldn't the enlarged QR code be a little bigger than this? Also, in my opinion, the 'Cancel' button can be removed and instead, clicking outside the box should close it.

app/src/main/java/org/fossasia/openevent/general/order/OrderDetailsFragment.kt
anhanh11001: yeah, I forgot to check for this

anhanh11001:
> Shouldn't the enlarged QR code be a little bigger than this? Also, in my opinion, the 'Cancel' button can be removed and instead, clicking outside the box should close it.

Ok, I'll try to find way to improve the size. Putting image view into a dialog was a bit trickier than I thought yesterday.

PR #1736, opened 2019-05-08
fix: Remove overlapping animation conflicts

### LINKED ISSUE #1734: Issue with returning animations and clicakble views
**Describe the bug**
Returning and entering animations for search date, location, and category fragment with search filter fragment are in the opposite way.
Full view is not clickable in search filter fragment, only text is clickable. 

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #1734


liveHarshit:
No animations are required while popping up to back stack because the back stack fragment already has returning animations.

PR #1754, opened 2019-05-12
feat: Update settings and profile fragment for non authenticated users

### LINKED ISSUE #1750: Cannot access settings and other options without authentication
**Describe the bug**
Split other options with the profile.

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #1750 
 For not logged in - 
<p align="center">
 
 
</p>
For logged in -
<p align="center">
 
 
</p>

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragment.kt
@@ -23,6 +23,7 @@ import kotlinx.android.synthetic.main.fragment_profile.view.logoutLL
 import kotlinx.android.synthetic.main.fragment_profile.view.manageEventsLL
 import kotlinx.android.synthetic.main.fragment_profile.view.settingsLL
 import kotlinx.android.synthetic.main.fragment_profile.view.ticketIssuesLL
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragment.kt
```diff
@@ -51,11 +52,19 @@ class ProfileFragment : Fragment() {
 
     override fun onStart() {
         super.onStart()
-        if (!profileViewModel.isLoggedIn()) {
-            redirectToLogin()
+        if (profileViewModel.isLoggedIn()) {
+            handleLayoutVisibility(true)
+        } else {
+            handleLayoutVisibility(false)
```
anhanh11001: > if (profileViewModel.isLoggedIn()) {
            handleLayoutVisibility(true)
        } else {
            handleLayoutVisibility(false)
>
This can be written as a single line like `handleLayoutVisibility(profileViewModel.isLoggedIn())`. I think it is cleaner

liveHarshit: Oh, thanks for pointing. :+1:

liveHarshit:
Updated

Issue #862, opened 2019-01-08
SearchView closes and data lost on rotation in search fragment.


kshitijsuri90:
**Describe the bug**

On rotation in search fragment while searchview is open, focus and data lost on rotation

**To Reproduce**

1. Go to the search fragment.
2. Open search option and enter text 
3. Rotate device.

**Expected behavior**

SearchView should remain in focus without loss of data

**Logs**

**Screenshots**

![search_rotation](https://user-images.githubusercontent.com/34534870/50852521-bf358c80-13a5-11e9-86ed-7def98228514.gif)

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |One Plus 5|
|Android Version|Android Pie 9.0.0  |

**Additional context**

@iamareebjamal @nikit19 Your views?

**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

kshitijsuri90:
Loss of data on rotation is becoming a common bug. Should we disable rotation like in Eventbrite?

iamareebjamal:
No. It's like killing the patient

vaibhav2203:
solved this issue, please review the commit.

kshitijsuri90:
Like I said you cannot just take up someone else's issue. @vaibhav2203 I mentioned that I am working on this (see above).

PR #1756, opened 2019-05-12
fix: Retain searchview state in SearchFragment

anhanh11001:
Detail:
- Retain search query in SearchFragment and require focus when making search on screen rotation

Fixes: #862

Screenshots for the change:
ezgif-4-41d82fe2dbf4

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
@@ -127,11 +127,19 @@ class SearchFragment : Fragment() {
         rootView.fabSearch.setOnClickListener {
             queryListener.onQueryTextSubmit(searchView.query.toString())
         }
+
+        if (searchViewModel.isQuerying) {
+            searchItem.expandActionView()
+            searchView.setQuery(searchViewModel.searchViewQuery, false)
+            searchView.clearFocus()
+        }
         super.onPrepareOptionsMenu(menu)
     }
 
     override fun onDestroyView() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
```diff
@@ -127,11 +128,22 @@ class SearchFragment : Fragment() {
         rootView.fabSearch.setOnClickListener {
             queryListener.onQueryTextSubmit(searchView.query.toString())
         }
+
+        if (searchViewModel.isQuerying) {
+            Handler().post {
```
adityastic: Never create an anonymous instance of handler. This will create a memory leak

iamareebjamal: Why

adityastic: This is a memory leak btw

anhanh11001: @addiegupta thanks for the warning, I will take a look at that. Initially I user this because there were some problems on rendering the view of the action bar. Will fix this

anhanh11001:
updated. please review @liveHarshit

app/src/main/java/org/fossasia/openevent/general/search/SearchFragment.kt
adityastic: @anhanh11001 😂😂 Wrong Aditya

Issue #1017, opened 2019-02-03
Provide option to  remove image, first name and last name to user.


shiv07tiwari:
**Describe the feature you'd like**
During SignUp, Image, first name and last name are optional for the user. So once a user sets an image, first name,last name, then he/she should have a feature to remove those fields from the account.

**Would you like to work on the issue?**
- [x] Yes
- [ ] No

theLimitBreaker:
@shiv07tiwari I think removing Profile picture only would be fine . Please confirm are you still working on this

PR #1757, opened 2019-05-12
feat: Add delete user photo feature

anhanh11001:
Detail:
- Update user photo by default image (encoded)

Fixes: #1017

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
@@ -14,6 +14,7 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.ContextCompat
+import androidx.core.graphics.drawable.toBitmap
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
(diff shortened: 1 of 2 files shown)


iamareebjamal:
But it won't be cleared on server?

anhanh11001:
@iamareebjamal You mean the previous user profile image? I think the server doesn't have the option to delete an image, only to upload it. By deleting user photo in issue #1017, I think it means just setting the profile image to a default image

iamareebjamal:
You can set it to null on the server

anhanh11001:
@iamareebjamal I think when sending a null image URL to the server, the current way server handles it is to ignore it and retain the previous profile image URL. I'm working with @uds5501 on this.

iamareebjamal:
@anhanh11001 Any update about the QR change you did. Is it compatible with the orga app?

anhanh11001:
Oh sorry for forgetting to finish it, I'm doing it now. Previously, I did take a look at it but the ScanActivity in orga app was blocked so I was confused but I forgot to ask you about that. I'll take a look again now and update to you later today.

anhanh11001:
@iamareebjamal I have checked it, it is handled correctly with "orderIdentifier-attendeeId"
save

ShridharGoel:
> Oh sorry for forgetting to finish it, I'm doing it now. Previously, I did take a look at it but the ScanActivity in orga app was blocked so I was confused but I forgot to ask you about that. I'll take a look again now and update to you later today.

Scanning is enabled in the Play Store build variant.

PR #1766, opened 2019-05-15
fix: Remove click listeners to remove memory leaks

anhanh11001:
Detail:
- The cause of the memory leak is because of the adapter of RecyclerView. If the adapter lives any longer than the RecyclerView does, the adapter is going to hold a reference to the RecyclerView which should have already gone out of memory.

Further description: https://stackoverflow.com/questions/35520946/leak-canary-recyclerview-leaking-madapter. 

To be honest, I don't completely understand the problem and the memory leak in here, but this fix does really improve the app as it is faster in my device and leak canary doesn't find any leaks. So please correct me if there is anything wrong or give me further explanations.

Fixes: #1780

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -558,6 +558,17 @@ class EventDetailsFragment : Fragment() {
         super.onPrepareOptionsMenu(menu)
     }
 
+    override fun onDestroyView() {
+        super.onDestroyView()
+        speakersAdapter.onSpeakerClick = null
+        sponsorsAdapter.onSponsorClick = null
+        sessionsAdapter.onSessionClick = null
+        similarEventsAdapter.apply {
+            onEventClick = null
+            onFavFabClick = null
+        }
+    }
+
     private fun loadTicketFragment() {
         val currency = Currency.getInstance(eventViewModel.event.value?.paymentCurrency ?: "USD").symbol
         findNavController(rootView).navigate(EventDetailsFragmentDirections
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventViewHolder.kt
```diff
@@ -53,11 +52,14 @@ class EventViewHolder(override val containerView: View) : RecyclerView.ViewHolde
         val startsAt = EventUtils.getEventDateTime(event.startsAt, event.timezone)
         val endsAt = EventUtils.getEventDateTime(event.endsAt, event.timezone)
 
+        containerView.shareFab.scaleType = CENTER
+        containerView.favoriteFab.scaleType = CENTER
```
ShridharGoel: How is this related with the issue?

anhanh11001: Oops. It's not. I just forgot to revert this

iamareebjamal:
Rather than doing this, just remove koin scoping please

anhanh11001:
I will remove Koin Scope. But should I keep the part where setting recyclerView adapter equals to null in other fragments in onDestroyView()? Leak Canary also find leaks in other fragments with the same adapter-recyclerView problem mentioned above.

iamareebjamal:
Honestly, that's a workaround, not a fix. We should try to find out the root cause of the leaks and fix it for the non koin adapters where leak canary is showing leaks. It may be because of nested viewholders or click listeners in adapters

anhanh11001:
@iamareebjamal please take the look at this article as it faces the same problem as we are and please give me some advice on which solution should I go for

https://medium.com/@yfujiki/tracing-simple-memory-leak-around-recyclerview-using-leakcanary-927460532d53

iamareebjamal:
I'll take a look over the weekend.

anhanh11001:
@iamareebjamal Have you taken a look at the blog post?

iamareebjamal:
Yes, the blog post is simply telling the problem. I'll analyse the app today to tell what can be the fix

iamareebjamal:
Memory leaks are due to click listeners. Please remove them onDestroyView

anhanh11001:
updated

Issue #1631, opened 2019-04-19
Create Call For Speaker


aggarwalpulkit596:
**Describe the feature you'd like**
Add a section for the call for speaker with an option to send a proposal along with speaker profile, session and track for logged in users.

**Would you like to work on the issue?**
- [x] Yes

aggarwalpulkit596:
@iamareebjamal need a suggestion 
https://open-event-api.herokuapp.com/#speakers-calls-get-speakers-call-for-an-event-get
this endpoint is authenticated, it makes sense as well as to be the speaker you need to login first to  submit proposal but it can be like when we click on button apply for speaker then check for authetication similar process is done for register button for event tickets

iamareebjamal:
Not needed for attendee app as of now

liveHarshit:
@iamareebjamal I think it should implement now. The frontend is also filtering events based on call for speakers and receiving speaker's proposal.

PR #1792, opened 2019-05-19
feat: Add call for speaker section

anhanh11001:
Detail:
- Set up menu action to get to Speakers Call Section
- I haven't implemented the submitting proposal feature so the button is not working now as I don't want to make a large PR. I'll open a new issue after this get merged to implement that.
- Putting Speakers Call inside menu item is just temporary. I'll open a new issue later to set up the tab layout so that every sections in EventDetailsFragment are placed logically and user-friendly.

Fixes: #1631 

Screenshots for the change:
              
ezgif-2-23f1dc967d2d


app/src/main/java/org/fossasia/openevent/general/OpenEventDatabase.kt
```diff
@@ -41,7 +43,7 @@ import org.fossasia.openevent.general.ticket.TicketIdConverter
 
 @Database(entities = [Event::class, User::class, SocialLink::class, Ticket::class, Attendee::class,
     EventTopic::class, Order::class, CustomForm::class, Speaker::class, SpeakerWithEvent::class, Sponsor::class,
-    SponsorWithEvent::class, Session::class], version = 6)
+    SponsorWithEvent::class, Session::class, SpeakersCall::class], version = 6)
```
liveHarshit: I think json file for version = 6 is not committed in development?

app/src/main/java/org/fossasia/openevent/general/speakercall/SpeakersCallViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+            mutableError.value = resource.getString(R.string.error_fetching_event_section_message,
+                resource.getString(R.string.speakers_call))
+            mutableProgress.value = false
+            return
+        }
+
+        compositeDisposable += eventService.getSpeakerCall(eventId)
+            .withDefaultSchedulers()
+            .doFinally {
+                mutableProgress.value = false
```
liveHarshit: IMO it's better if you set default visibility of progress to GONE and set the mutableProgress value to true onSubscribe.

app/src/main/java/org/fossasia/openevent/general/OpenEventDatabase.kt
anhanh11001: Yes, it isn't

anhanh11001:
updated

Issue #1692, opened 2019-04-29
Unable to mark event as favourite through toolbar in event details 


ShridharGoel:
**Describe the bug**
Unable to mark event as favourite through toolbar in event details 

**GIF**
![ezgif com-video-to-gif (67)](https://user-images.githubusercontent.com/35566748/56923374-ff9d4f80-6ae7-11e9-8c96-53b05ed1404a.gif)

**Would you like to work on the issue?**
Open to all.

liveHarshit:
@ShridharGoel I'm not able to reproduce this issue. The issue is that event marked as favorite but recycle view is not updating after going back.

ShridharGoel:
For me, the like icon is changing back to unliked state as in the GIF.

ShridharGoel:
Will try again on latest commits

ShridharGoel:
The issue is still there, please check @liveHarshit @anhanh11001

anhanh11001:
Yes, I can still reproduce it, i'll take a look at it

anhanh11001:
@ShridharGoel  Can you check if this issue is still reproducible. I think it is fixed now by #1707. I guess the reason was when calling this method to load the flowable from the database, we missed `.distinctUntilChanged()` so the UI constantly keeps updating (like after every 1 second) by following the LiveData, which led to weird UI behaviour in EventDetailsFragment

Now:
```
fun loadEvent(id: Long) {
        if (id == -1L) {
            mutablePopMessage.value = resource.getString(R.string.error_fetching_event_message)
            return
        }
        compositeDisposable += eventService.getEvent(id)
            .withDefaultSchedulers()
            .distinctUntilChanged() // This was missed before
            .doOnSubscribe {
                mutableProgress.value = true
            }.doFinally {
                mutableProgress.value = false
            }.subscribe({
                mutableEvent.value = it
            }, {
                Timber.e(it, "Error fetching event %d", id)
                mutablePopMessage.value = resource.getString(R.string.error_fetching_event_message)
            })
    }
```

ShridharGoel:
Fixed.

PR #1793, opened 2019-05-19
fix: liking fail in EventDetailsFragment

anhanh11001:
Detail:
- Change the database call for event from flowable to single object.

The main reason why clicking like behaviour is unexpected is because Flowable object keep emitting event for the ViewModel. Therefore, LiveData object about the event inside the ViewModel keeps updating constantly, and which leads to weird behavior inside this fragment.

Likes icon was updated because new event object is emitted so that we have to reset all the UI, not because we directly change the likes icon.

Fixes: #1692

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDao.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDao.kt
@@ -22,7 +22,7 @@ interface EventDao {
     fun getAllEvents(): Flowable<List<Event>>
 
     @Query("SELECT * from Event WHERE id = :id")
-    fun getEvent(id: Long): Flowable<Event>
+    fun getEvent(id: Long): Single<Event>
 
     @Query("SELECT * FROM event WHERE id = :eventId")
     fun getEventById(eventId: Long): Single<Event>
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventDao.kt
```diff
@@ -22,7 +22,7 @@ interface EventDao {
     fun getAllEvents(): Flowable<List<Event>>
 
     @Query("SELECT * from Event WHERE id = :id")
-    fun getEvent(id: Long): Flowable<Event>
+    fun getEvent(id: Long): Single<Event>
```
iamareebjamal: Please revert. This breaks live update of the event

iamareebjamal:
Solution is to use correct query for the like status. Not to change Flowable to Single

app/src/main/java/org/fossasia/openevent/general/event/EventDao.kt
anhanh11001: what do you mean by live update? I haven't understood why would we want a continuous stream of the same data

iamareebjamal: You don't get the same data. You get it if something in DB has changed

iamareebjamal: https://en.wikipedia.org/wiki/Single_source_of_truth

iamareebjamal: https://codelabs.developers.google.com/codelabs/android-persistence/#5

Issue #1751, opened 2019-05-11
Add authentication with facebook and google


liveHarshit:
**Would you like to work on the issue?**

- [x] Yes
- [ ] No
- Other:

mariobehling:
Out of scope currently.

PR #1794, opened 2019-05-20
feat: Add authentication with facebook

liveHarshit:
Part of #1751 
- Collect user data i.e. email, first name, last name using facebook api.
- Check if the email is already registered with Eventyay.
- If yes then redirect to login and display msg about email is already registered with us.
- If not then show a password alert dialog and ask the user to provide password and sign up.
- Log in automatically and redirect to back stack
- Declear api keys as environmental variable
- Hide login button if api key is not provided
- Move logic to view model

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,6 +10,8 @@ apply plugin: 'androidx.navigation.safeargs.kotlin'
 
 def STRIPE_API_TOKEN = System.getenv('STRIPE_API_TOKEN') ?: "YOUR_API_KEY"
 def MAPBOX_KEY = System.getenv('MAPBOX_KEY') ?: "pk.eyJ1IjoiYW5nbWFzMSIsImEiOiJjanNqZDd0N2YxN2Q5NDNuNTBiaGt6eHZqIn0.BCrxjW6rP_OuOuGtbhVEQg"
+def FB_APP_ID = System.getenv('FB_APP_ID') ?: "NOT_PROVIDED"
+def FB_LOGIN_PROTOCOL = System.getenv("FB_LOGIN_PROTOCOL") ?: "NOT_PROVIDED"
 
 android {
     dataBinding {
@@ -26,7 +28,8 @@ android {
         vectorDrawables.useSupportLibrary = true
         multiDexEnabled true
         manifestPlaceholders = [
-                STRIPE_API_TOKEN: STRIPE_API_TOKEN
+                STRIPE_API_TOKEN: STRIPE_API_TOKEN,
+                FB_LOGIN_PROTOCOL: FB_LOGIN_PROTOCOL
         ]
         javaCompileOptions {
             annotationProcessorOptions {
@@ -41,10 +44,12 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
             buildConfigField "String", "DEFAULT_BASE_URL", '"https://api.eventyay.com/v1/"'
             buildConfigField "String", "MAPBOX_KEY", '"'+MAPBOX_KEY+'"'
+            resValue "string", "FB_APP_ID", '"'+FB_APP_ID+'"'
         }
         debug {
             buildConfigField "String", "DEFAULT_BASE_URL", '"https://open-event-api-dev.herokuapp.com/v1/"'
             buildConfigField "String", "MAPBOX_KEY", '"'+MAPBOX_KEY+'"'
+            resValue "string", "FB_APP_ID", '"'+FB_APP_ID+'"'
         }
     }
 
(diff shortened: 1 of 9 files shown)


liveHarshit:
@anhanh11001 Please review

app/build.gradle
```diff
@@ -187,6 +187,10 @@ dependencies {
     //LeakCanary
     debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.0-alpha-1'
 
+    //Facebook authentication
+    implementation 'com.facebook.android:facebook-login:[5,6)'
```
iamareebjamal: Use exact versioning

app/src/main/java/org/fossasia/openevent/general/auth/AuthFragment.kt
```diff
@@ -20,16 +32,33 @@ import org.fossasia.openevent.general.utils.Utils.hideSoftKeyboard
...
+private const val LAST_NAME = "last_name"
 
 class AuthFragment : Fragment() {
     private lateinit var rootView: View
     private val authViewModel by viewModel<AuthViewModel>()
     private val safeArgs: AuthFragmentArgs by navArgs()
     private val smartAuthViewModel by sharedViewModel<SmartAuthViewModel>()
+    private val callbackManager = CallbackManager.Factory.create()
+    private lateinit var signUp: SignUp
+    private var fbLogIn = false
```
anhanh11001: I suggest keeping parameter about the logic inside viewModel

anhanh11001:
Can you check the implementaition as well because I couldn't log in with my facebook
ezgif-4-cf4ded456e3d

app/src/main/res/values/strings.xml
```diff
@@ -357,4 +359,8 @@
     <string name="call_for_speakers">Call For Speakers</string>
     <string name="submit_proposal">Submit Proposal</string>
 
+    <!--Facebook auth-->
+    <string name="facebook_app_id">{Facebook app id}</string>
+    <string name="fb_login_protocol_scheme">{Login protocol scheme}</string>
```
liveHarshit: @anhanh11001 I have removed the API key. Please add the following and test it - 
```xml
<string name="facebook_app_id">1240102032819527</string>
<string name="fb_login_protocol_scheme">fb1240102032819527</string>
```

app/src/main/java/org/fossasia/openevent/general/auth/AuthFragment.kt
liveHarshit: Here we need to register `callbackManager` with the fragment and handle activity results so it will complex to use view model. I will try it. :+1:

app/src/main/res/values/strings.xml
iamareebjamal: This is not configurable at all and will break the release process. You don't expect us to pull the code, replace the value and build the app each time right?

iamareebjamal: Also, if the values are not present, the button should not be shown

liveHarshit: > This is not configurable at all and will break the release process. You don't expect us to pull the code, replace the value and build the app each time right?

Yes, I'll add these in the manifest and define them in gradle.build.

liveHarshit:
@anhanh11001 @iamareebjamal Please review

Updated - 
- Declear api keys as environmental variable
- Hide login button if api key is not provided
- Move logic to view model

> it has the same height as our current button

I have tried to increase the size but not working as it is a default button by facebook SDK.

app/build.gradle
```diff
@@ -10,6 +10,8 @@ apply plugin: 'androidx.navigation.safeargs.kotlin'
 
 def STRIPE_API_TOKEN = System.getenv('STRIPE_API_TOKEN') ?: "YOUR_API_KEY"
 def MAPBOX_KEY = System.getenv('MAPBOX_KEY') ?: "pk.eyJ1IjoiYW5nbWFzMSIsImEiOiJjanNqZDd0N2YxN2Q5NDNuNTBiaGt6eHZqIn0.BCrxjW6rP_OuOuGtbhVEQg"
+def FB_APP_ID = System.getenv('FB_APP_ID') ?: "NOT_PROVIDED"
+def FB_LOGIN_PROTOCOL = System.getenv("FB_LOGIN_PROTOCOL") ?: "NOT_PROVIDED"
```
iamareebjamal: What should be passed here?

liveHarshit: I have declared these as environmental can be set during release build, same as strip key.

liveHarshit:
> This is not the correct way to use social login. Password will not be asked in any social auth scenario

Yes I have also thought that but not able to sign up on Eventyay without a password. Please suggest another way?

iamareebjamal:
There's already social auth implemented on backend. Check the docs

liveHarshit:
> There's already social auth implemented on backend. Check the docs

Sorry to say, but I did not find any endpoint in the docs - http://dev.eventyay.com/api/v1

iamareebjamal:
It's not in docs. But it's implemented. Nevertheless, this is low priority and the implementation will not require any input from user

liveHarshit:
https://github.com/fossasia/open-event-server/issues/4526#issuecomment-478284564

mariobehling:
As this is out of scope of the current version, I am closing this.

Issue #1798, opened 2019-05-20
Event not loaded on EventDetailsFragment 


anhanh11001:
**Describe the bug**
My previous PR #1740 made a mistake on calling viewModel.loadEvent() function only when there is internet connection while the event is loaded from the database, not the api.
This also leads to a problem on shared element transition.

**Would you like to work on the issue?**
Yes

PR #1799, opened 2019-05-20
fix: Load event detail UI immediately on EventDetailsFragment

anhanh11001:
Detail:
- Previously, loadEvent() only called only when there is internet connection. Now it is loaded immediately when rootView is created.
- Delete postpone on shared element transition as sometimes it takes time to load the cover image. It is better to just make the transition with the default image if the cover image is not loaded.

Fixes: #1798 #1805

@liveHarshit can you test this implementation to see if there is any problem with shared element transition with your device

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -136,6 +136,12 @@ class EventDetailsFragment : Fragment() {
         setToolbar(activity)
         setHasOptionsMenu(true)
 
+        val currentEvent = eventViewModel.event.value
+        if (currentEvent != null)
+            loadEvent(currentEvent)
+        else
+            eventViewModel.loadEvent(safeArgs.eventId)
+
         rootView.feedbackRv.layoutManager = LinearLayoutManager(context)
         rootView.feedbackRv.adapter = feedbackAdapter
 
@@ -379,13 +385,6 @@ class EventDetailsFragment : Fragment() {
             .nonNull()
             .observe(this, Observer { isConnected ->
                 if (isConnected) {
-                    showEventErrorScreen(false)
-                    val currentEvent = eventViewModel.event.value
-                    if (currentEvent == null)
-                        eventViewModel.loadEvent(safeArgs.eventId)
-                    else
-                        loadEvent(currentEvent)
-
                     val currentFeedback = eventViewModel.eventFeedback.value
                     if (currentFeedback == null) {
                         eventViewModel.fetchEventFeedback(safeArgs.eventId)
@@ -435,12 +434,6 @@ class EventDetailsFragment : Fragment() {
                         rootView.socialLinkContainer.visibility =
                             if (currentSocialLinks.isEmpty()) View.GONE else View.VISIBLE
                     }
-                } else {
-                    val currentEvent = eventViewModel.event.value
-                    if (currentEvent == null)
-                        showEventErrorScreen(true)
-                    else
-                        loadEvent(currentEvent)
                 }
             })
 


app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -297,15 +301,7 @@ class EventDetailsFragment : Fragment() {
...
-            .into(rootView.eventImage, object : Callback {
-                override fun onSuccess() {
-                    startPostponedEnterTransition()
-                }
-
-                override fun onError(e: Exception?) {
-                    startPostponedEnterTransition()
-                }
-            })
+            .into(rootView.eventImage)
```
liveHarshit: Can we pass the event image from Events fragment to event details fragment, so there will be no delay for loading image using URL and Picasso?

liveHarshit:
Also mention Fixes #1805

liveHarshit:
Search is not working, the delay is fixed :+1:

anhanh11001:
And by search you mean this fragment right?


Can you describe the behaviour or attach a GIF because I couldn't reproduce the bug and my device is working fine. @liveHarshit

liveHarshit:
> or attach a GIF


ezgif com-video-to-gif (7)

anhanh11001:
@liveHarshit maybe your device info as well..?

liveHarshit:
@anhanh11001 
Mi Max 2 (Android 7.1.1 API 25)

Working fine with Nexus 5X API Q

liveHarshit:
Also same for issue #1804 
Something wrong with my device... :sweat_smile:

liveHarshit:
But this PR is ready to merge. :+1: It is not related to search.

anhanh11001:
I can reproduce it with device in older version as well

liveHarshit:
Oh, see what happens when you go to likes section in the old device.

anhanh11001:
also the same bug as #1804

liveHarshit:
@anhanh11001 I have tested it. These bugs and multiple crashes are due to shared element transition for device version 7 or before.

liveHarshit:
So what should we do, solve these bugs one by one or change the method for shared element transition? Because for the Eventbrite app, the transition is working fine for old devices too.

anhanh11001:
@liveHarshit I'm not sure. From the log I cannot trace back to any part of the code, so I guess this might be some problem due to Navigation component that relate to shared element transition. 
I don't know why Eventbrite is working fine, maybe because they set that up in an older way. For now, I have fixed in this PR by only making shared element transition for Android API 26 or upper.

app/build.gradle
```diff
@@ -43,7 +43,9 @@ android {
             buildConfigField "String", "MAPBOX_KEY", '"'+MAPBOX_KEY+'"'
         }
         debug {
-            buildConfigField "String", "DEFAULT_BASE_URL", '"https://open-event-api-dev.herokuapp.com/v1/"'
+//            buildConfigField "String", "DEFAULT_BASE_URL", '"https://open-event-api-dev.herokuapp.com/v1/"'
```
liveHarshit: Wrong file committed

liveHarshit:
> PR by only making shared element transition for Android API 26 or upper.

Good for now.

app/build.gradle
anhanh11001: oh right, but I'm gonna split this PR now, one for loadEvent() function, the other for sharedElementTransition

iamareebjamal:
> PR by only making shared element transition for Android API 26 or upper

But why? THey are supported since Lollipop

anhanh11001:
To be honest, I don't know if it is because of the way I set up shared element is wrong or because there is some problem with navigation component. I received a log about NullPointerException but it doesn't trace back to any part of the code. The fix in #1807 is just a temporary fix to avoid the crash in older version (Harshit and I have tested and the current implementation is working fine with API 26 or higher).

Issue #1803, opened 2019-05-20
App crashes on navigating back from AuthFragment


anhanh11001:
**Describe the bug**
App crashes on navigating back from AuthFragment

**To Reproduce**
1. Go to the app without logged in
2. Move to OrderDetailsFragment -> Will be redirect to LogInFragment 
3. Click back button -> App crashes

**Would you like to work on the issue?**
Yes

PR #1806, opened 2019-05-21
fix: crash on coming back from AuthFragment

anhanh11001:
Detail:
- popBackstack() changed into navigating from from action pop backstack in navigation graph. This might be a problem from Navigation Component as both methods are pretty much the same but the latter one doesn't produce the crashes so this is just a workaround.
Please give an advice if a better solution can be made

Fixes: #1803

Screenshots for the change:
ezgif-2-ad24393b3574


liveHarshit:
> * Please give an advice if a better solution can be made

I Have done the same for my last PR. The app was crashing on the welcome fragment.

Issue #1811, opened 2019-05-21
Set up send proposal for Call for Speakers


anhanh11001:
**Describe the feature you'd like**
Currently, Send Proposal Button in Call for Speakers is not implemented.

**Would you like to work on the issue?**
Yes

PR #1816, opened 2019-05-24
feat: Add create/load speaker profile for user

anhanh11001:
Details:
- Create/Load speaker profile for proposal
- I'll finish the rest of #1811 (implementing create proposal after this PR is finished and merged as this is already a large PR)

Fixes: #1811

Screenshots for the change:
       

ezgif-4-94cf806dc384 
 ezgif-4-6315722607c0


anhanh11001:
@nikit19 Yes, I understand it, and both current WIP PR I'm sending take me more time then I expected, but I'll make sure I finish them as soon as possible, hopefully around today evening. I sent them to let the mentors and other contributors know that I'm still making some progress

anhanh11001:
@nikit19 Do you know how to make a filter in making GET request. I'm trying to take a speaker under event and user by this:
```
@GET("users/{userId}/speakers?include=event,user&fields[event]=id&fields[user]=id")
    fun getSpeakerUnderUserAndEvent
        (@Path("userId") userId: Long, @Query("event", encoded = true) eventId: EventId): Single<Speaker>
```
or
```
@GET("users/{userId}/speakers?include=event,user&fields[event]=id&fields[user]=id")
    fun getSpeakerUnderUserAndEvent
        (@Path("userId") userId: Long, @Query("event", encoded = true) eventId: Long): Single<Speaker>
```
But both of this return a list of speakers under user without filtering the event ID given.

iamareebjamal:
There's no filter applied and the query itself is wrong. Why are you querying `users/{user_id}`?

anhanh11001:
In the frontend, the information about my speaker is stored in each event when I try to send a proposal. If not, then I would have to create a new speaker. 



I'm trying to do the same thing but I don't know how to save the current speaker detail. So I was trying to create a speaker under a user and event, save that in the server and then query it to check whether I have an existing user or not.

I query `users/{user_id}` as I follow this from the documentation. This does return a list of speakers
Capture

iamareebjamal:
But you haven't applied any filter

anhanh11001:
So putting this `@Query("event", encoded = true) eventId: EventId` in the function is not correct way to filter.. . So I'll spend time on reading about the filter part and I will update if I still have problems

iamareebjamal:
`users/{userId}/speakers?include=event,user&fields[event]=id&fields[user]=id`

There is no filter in the query. Adding event in the query will just make it 

`users/33/speakers?include=event,user&fields[event]=id&fields[user]=id&event=234`

Also you are using encoded EventId and not an int, which will result in God knows what

How filters work if shown in Search section of the app. Please take a look at that

anhanh11001:
Updated: Description above

app/src/main/java/org/fossasia/openevent/general/speakers/Speaker.kt
```diff
@@ -15,24 +19,30 @@ data class Speaker(
...
+    val facebook: String? = null,
+    val linkedin: String? = null,
+    val github: String? = null,
+    val isFeatured: Boolean = false,
+    @ColumnInfo(index = true)
+    @Relationship("event")
+    var event: EventId? = null,
+    @ColumnInfo(index = true)
+    @Relationship("user")
+    var user: UserId? = null
```
liveHarshit: These all should be value not variables, right?

anhanh11001:
Updated, I have used progress dialog, it looks better; enable email editing (at first I thought the frontend set is as disabled); turned variables to values

app/src/main/java/org/fossasia/openevent/general/speakercall/SpeakersCallViewModel.kt
```diff
@@ -6,30 +6,91 @@ import androidx.lifecycle.ViewModel
...
+                |       'op':'has',
+                |       'val': {
+                |           'name': 'identifier',
+                |           'op': 'eq',
+                |           'val': '$eventIdentifier'
+                |       }
+                |    }, {
+                |       'name':'email',
+                |       'op':'eq',
+                |       'val':'${user.email}'
```
iamareebjamal: Why is the email thingy needed?

iamareebjamal:
I think you can remove the email from the filter. Rest is fine

app/src/main/java/org/fossasia/openevent/general/speakercall/SpeakersCallViewModel.kt
anhanh11001: I followed the implementation in the frontend. It seems like the server allow a user to create multiple speaker's profiles, but only allow one email per profile. 

dfs

PR #1835, opened 2019-05-27
feat: Add filter for sessions and speakers

### LINKED ISSUE #1825: Add search filters on event
- [x] Sessions and Speakers
- [ ] Call for speakers

- [x] Yes

liveHarshit:
Part of #1825

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/SearchFilterFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/SearchFilterFragment.kt
@@ -9,10 +9,12 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.RadioButton
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.Observer
 import androidx.navigation.Navigation.findNavController
 import androidx.navigation.fragment.navArgs
 import kotlinx.android.synthetic.main.fragment_search_filter.view.dateRadioButton
 import kotlinx.android.synthetic.main.fragment_search_filter.view.freeStuffCheckBox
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/openevent/general/search/SearchFilterFragment.kt
```diff
@@ -33,6 +34,7 @@ class SearchFilterFragment : Fragment() {
     private val searchViewModel by viewModel<SearchViewModel>()
     private val safeArgs: SearchFilterFragmentArgs by navArgs()
     private lateinit var sortBy: String
+    private var isSessionsAndSpeakersChecked = false
```
anhanh11001: Normally, I would put `isSessionsAndSpeakersChecked` inside ViewModel instead of Fragment, just to be sure that it survives screen rotation

liveHarshit:
@anhanh11001 yes it is not a filter for call for speakers, currently I did the same as frontend. Because there is no value in event type to filter call for speaker events, Need to find another way for it.

app/src/main/java/org/fossasia/openevent/general/search/SearchFilterFragment.kt
liveHarshit: Okay got it. I will also move other Variables inside view models.

anhanh11001:
Please resolve conflicts, rest is good

liveHarshit:
> Please resolve conflicts, rest is good

Done

liveHarshit:
@iamareebjamal Please review.

PR #1851, opened 2019-05-29
feat: Show date and time wise check-in

### LINKED ISSUE #1845: Show date wise check
Display multiple check-in time and dates.

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #1845

Screenshots for the change:
<p align="center">
 
 
</p>

### DIFF
--- a/app/schemas/org.fossasia.openevent.general.OpenEventDatabase/6.json
+++ b/app/schemas/org.fossasia.openevent.general.OpenEventDatabase/6.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 6,
-    "identityHash": "24ed50935955154c79491815778897fb",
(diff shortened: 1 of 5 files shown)


anhanh11001:
layout id/checkInLayout and layout id/notCheckInLayout is the same right? So instead of toggling the visibility of these layouts, why don't you make it a single layout and change the content (text, circle color) inside it

liveHarshit:
> layout id/checkInLayout and layout id/notCheckInLayout is the same right? So instead of toggling the visibility of these layouts, why don't you make it a single layout and change the content (text, circle color) inside it

The code will get complicated because in that case, I need to change text color, set background tint and change text programmatically for each condition. So, IMO it's better to set design in layout resources file and make the code simple.

anhanh11001:
Well, that makes sense to me

Issue #1857, opened 2019-05-30
Add checking required fields on registering tickets


anhanh11001:
**Describe the feature you'd like**
Add checking required fields on registering tickets and show error message when clicking register button.

**Would you like to work on the issue?**
Yes

PR #1859, opened 2019-05-30
feat: Add checking required fields

anhanh11001:
Details:
- Add checking required fields
- Add hint for required fields

Fixes: #1857

Screenshots for the change:


anhanh11001:
@liveHarshit I think so, which part of it do you find not necessary. For billing information, I only check it when the checkbox is ticked

liveHarshit:
> I only check it when the checkbox is ticked

Okay then. :+1:

Issue #1830, opened 2019-05-27
Stripe implementation is not finished


anhanh11001:
**Describe the bug**
In createOrder() function inside AttendeesViewModel,

```
if (it.paymentMode == resource.getString(R.string.free)) {
                        confirmOrder = ConfirmOrder(it.id.toString(), "completed")
                        confirmOrderStatus(it.identifier.toString(), confirmOrder)
                    } else mutableMessage.value = resource.getString(R.string.order_success_message)
```
There is no order confirm for paid order so the paid order is not created. This is not tested as well.

**Would you like to work on the issue?**
Yes

anhanh11001:
From the documentation and implementation in the frontend, I understand there are 5 payment options: Online - Paypal, Stripe and Offline - On site, cheque, telegraphic transfer. 
<img src="https://i.ibb.co/HgtH9hR/Capture.png" alt="Capture" border="0">
<img src="https://i.ibb.co/6D73vKH/payment-1.png" alt="payment-1" border="0">

So for offline payments method, I'll display instruction information. I guess for onsite,cheque and telegraphic transfer, we can't do it in the app. For paypal, I'll display some message indicating that user should forward to website to make Paypal payment as it is not integrated in the app(server). For Stripe, I'm still working on it to make it work. Please let me know if there are any information I need to know about payment

There is also a part about tax information, should I care about that for now?

iamareebjamal:
You need to fetch event from server to see what payment methods are supported for the event. A person may have disabled a particular option, so it should not be shown.

Other than that, you are correct.

PR #1863, opened 2019-05-30
feat: Implement Stripe and charging

anhanh11001:
Details:
- Remove card type picker as this is a unnecessary component
- Add Stripe implementation to the flow of ordering ticket

Flow of ordering paid ticket with Stripe:
-> Create order -> Send token -> Charge order -> Confirm order

Fixes: #1830


anhanh11001:
@iamareebjamal how should I test whether I can create paid order successfully with Stripe. I have tried creating my own Stripe account, get test Publishable key, put that key into the build.gradle file, use test card from stripe. But I get this error during charging

PR #1889, opened 2019-06-03
fix: Sign Up not working

### LINKED ISSUE #1870: Authentication fragments e-mail validation not working.
**Describe the bug**

In the main sign up page even if entering correct email address, the validator is throwing error stating invalid email address. 
**To Reproduce**

    1.Go to signup.
    2.Enter correct email ID.
    3.Next complete Signup page.
    4.Enter correct email ID.
    5.It will not be accepted.

**Expected behavior**

Form should proceed after correct validation. 

**Screenshots**
Testing

- [x] Yes
- [ ] No

liveHarshit:
Fixes #1870 

- According to the server implementation, password length should be 8 characters.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/SignUpFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/SignUpFragment.kt
@@ -48,6 +48,8 @@ import org.fossasia.openevent.general.ticket.TICKETS_FRAGMENT
 import org.jetbrains.anko.design.longSnackbar
 import org.jetbrains.anko.design.snackbar
 
+private const val MINIMUM_PASSWORD_LENGTH = 8
+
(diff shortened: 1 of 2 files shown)


ShridharGoel:
Shouldn't the error message be changed since it specifically says that the email is invalid while error 422 can be for other reasons too?

liveHarshit:
> Shouldn't the error message be changed since it specifically says that the email is invalid while error 422 can be for other reasons too?

Yes. @nikit19 Where can I find docs about these error messages, because these are not in the server docs?

anhanh11001:
So what was the reason for sign up not working properly (just because of short password?)

vartult:
> So what was the reason for sign up not working properly (just because of short password?)

Yes the required minimum length for password is 8.

Issue #1897, opened 2019-06-04
EventDetailsFragment display unaccepted sessions/speakers


anhanh11001:
**Describe the bug**
Sessions/Speakers should only be displayed when they are accepted by the Organizer
The same for public/private events/speakerscall

**Would you like to work on the issue?**
Yes

liveHarshit:
Parent issue?

anhanh11001:
I removed child-issue. I mistakenly added that tag

PR #1902, opened 2019-06-04
fix: Display valid information about sessions, speakers, call for speakers

anhanh11001:
Display only:
- featured speakers
- accepted sessions
- public call for speaker

Fixes: #1897


iamareebjamal:
Open an issue on server. It should show only accepted things by default

anhanh11001:
@iamareebjamal But then I think the orga app can't receive the full list of pending speakers/sessions?

iamareebjamal:
Onbviously, an organizer can and will be able to see all. Attendee shouldn't be able to

anhanh11001:
oh never mind, I was overthinking of an another problem. I'm opening an issue now

anhanh11001:
As the session and speakers call are handled by the server, I only made the filter for speakers

iamareebjamal:
It'd be better if there's no featured speaker, fall back to all speakers

PR #1918, opened 2019-06-06
fix: Revalidate profile from API

### LINKED ISSUE #1904: Fetch profile from API and revalidate from database
Add refresh button to retrieve user information from the server instead of the database.
Whenever profile information updated from server side it will not retrieve in the app.

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #1904

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
@@ -76,6 +76,14 @@ class AuthService(
                 }
     }
 
+    fun syncProfile(id: Long = authHolder.getId()): Single<User> {
+        return authApi.getProfile(id)
+            .map {
+                userDao.insertUser(it)
+                it
+            }
+    }
+
     fun sendResetPasswordEmail(email: String): Single<RequestTokenResponse> {
         val requestToken = RequestToken(Email(email))
         return authApi.requestToken(requestToken)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -76,6 +76,14 @@ class AuthService(
                 }
     }
 
+    fun updateProfile(id: Long = authHolder.getId()): Single<User> {
```
iamareebjamal: Where is it updating the profile?

liveHarshit: Inserting in the database and updating in the profile fragment without showing progress.

iamareebjamal: That's fetching/saving/syncing

Updating is if you sent come data on server to update/

liveHarshit: Not getting it. Mean retrieving information from the server and updating the same on the server.

liveHarshit: > come data

Data coming from API?

iamareebjamal: Updated

liveHarshit: But here we are updating User name only in edit profile fragment which is already updating on server. And now syncing the data is required because profile verification is working from the server side.

iamareebjamal: Why is the name of method updateProfile?

liveHarshit: Oh got it. It shouldn't. It can be sync or fetch profile.

liveHarshit: Updated

Issue #1895, opened 2019-06-04
App crashes on clicking SimilarEvent in OrderCompletedFragment


anhanh11001:
**Describe the bug**
App crashes on clicking SimilarEvent in OrderCompletedFragment

**Logs**
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.eventyay.attendee, PID: 17305
    java.lang.NullPointerException: Attempt to invoke interface method 'boolean java.util.Collection.contains(java.lang.Object)' on a null object reference
        at androidx.collection.MapCollections.retainAllHelper(MapCollections.java:481)
        at androidx.collection.ArrayMap.retainAll(ArrayMap.java:164)
        at androidx.fragment.app.FragmentTransition.captureOutSharedElements(FragmentTransition.java:803)
        at androidx.fragment.app.FragmentTransition.configureSharedElementsReordered(FragmentTransition.java:575)
        at androidx.fragment.app.FragmentTransition.configureTransitionsReordered(FragmentTransition.java:235)
        at androidx.fragment.app.FragmentTransition.startTransitions(FragmentTransition.java:135)
        at androidx.fragment.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:1869)
        at androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManagerImpl.java:1817)
        at androidx.fragment.app.FragmentManagerImpl.execPendingActions(FragmentManagerImpl.java:1717)
        at androidx.fragment.app.FragmentManagerImpl$2.run(FragmentManagerImpl.java:150)
        at android.os.Handler.handleCallback(Handler.java:808)
        at android.os.Handler.dispatchMessage(Handler.java:101)
        at android.os.Looper.loop(Looper.java:166)
        at android.app.ActivityThread.main(ActivityThread.java:7529)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:245)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:921)
```
**Screenshots**
<img src="https://i.ibb.co/yypvJDv/ezgif-2-6a2046d3f6cc.gif" alt="ezgif-2-6a2046d3f6cc" border="0">

**Would you like to work on the issue?**
Yes

liveHarshit:
Not crashing on my device. You android version?

anhanh11001:
[...]

PR #1919, opened 2019-06-06
fix: Avoid crash on OrderCompleted

anhanh11001:
Details:
- Shorter implementation of shared element transition
- Remove pop fragments in the stack (followed Eventbrite navigation). Initially, the crash was due to shared element transition with transition containing pop fragments in backstack (this is what I guess, from the log I couldn't really find out the real problem). I checked the implementation in Eventbrite and they don't pop fragment in backstack for this, so I remove unnecessary options on navigation
```
<action
            android:id="@+id/action_order_completed_to_event_detail"
            app:destination="@id/eventDetailsFragment"
            app:popUpTo="@id/eventsFragment" // removed
            app:popUpToInclusive="false" /> // removed
```

Fixes: #1895

### DIFF
--- a/app/src/main/res/navigation/navigation_graph.xml
+++ b/app/src/main/res/navigation/navigation_graph.xml
@@ -664,8 +664,10 @@
         <action
             android:id="@+id/action_order_completed_to_event_detail"
             app:destination="@id/eventDetailsFragment"
-            app:popUpTo="@id/eventsFragment"
-            app:popUpToInclusive="false" />
+            app:popEnterAnim="@anim/slide_in_left"
+            app:popExitAnim="@anim/slide_out_right"
+            app:enterAnim="@anim/slide_in_right"
+            app:exitAnim="@anim/slide_out_left"/>
 
         <argument
             android:name="eventId"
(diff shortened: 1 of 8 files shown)


app/src/main/res/navigation/navigation_graph.xml
```diff
@@ -664,8 +664,10 @@
         <action
             android:id="@+id/action_order_completed_to_event_detail"
             app:destination="@id/eventDetailsFragment"
-            app:popUpTo="@id/eventsFragment"
-            app:popUpToInclusive="false" />
+            app:popEnterAnim="@anim/slide_in_left"
```
liveHarshit: What happens if you press back, it should not redirect to order completed fragment again?

anhanh11001: No it still redirect to back to that one, please check this one
ezgif-4-5b4e4155cde4

liveHarshit: Cool.

liveHarshit: Sorry I have not checked Eventbrite completely. :smile:

PR #1940, opened 2019-06-10
feat: Display tag for already registered tickets

### LINKED ISSUE #1875: Diaplay tag for already registered tickets
Display tag for already registered tickets in Event details fragment.

- [x] Yes
- [ ] No
- Other:

liveHarshit:
Fixes #1875 
<p align="center">
 
</p>

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -51,6 +51,7 @@ import kotlinx.android.synthetic.main.content_event.view.socialLinksRecycler
 import kotlinx.android.synthetic.main.content_event.view.socialLinkContainer
 import kotlinx.android.synthetic.main.content_event.view.similarEventsRecycler
 import kotlinx.android.synthetic.main.content_event.view.similarEventsContainer
+import kotlinx.android.synthetic.main.content_event.view.alreadyRegisteredLayout
 import kotlinx.android.synthetic.main.fragment_event.view.buttonTickets
 import kotlinx.android.synthetic.main.fragment_event.view.eventErrorCard
 import kotlinx.android.synthetic.main.fragment_event.view.container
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventDetailsViewModel.kt
```diff
@@ -241,6 +246,16 @@ class EventDetailsViewModel(
             })
     }
 
+    fun loadOrders() {
+        compositeDisposable += orderService.getOrdersOfUser(getId())
```
anhanh11001: I think there should be a check statement if user is loggedIn() or not in here so that we won't waste a network call

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -154,6 +156,30 @@ class EventDetailsFragment : Fragment() {
             loadTicketFragment()
         }
 
+        eventViewModel.loadOrders()
+        eventViewModel.orders
```
anhanh11001: You can check if eventViewModel.orders.value == null before calling eventViewModel.loadOrders(). If it is not null, then you can call the method below immediately without making the network call again. This should prevent making the API call again after screen rotation or coming back to this fragment from backstack. 
```
it.forEach { order ->
                    if (order.event?.id == safeArgs.eventId) {
                        rootView.alreadyRegisteredLayout.visibility = View.VISIBLE
                        orderId = order.id
                        rootView.alreadyRegisteredLayout.setOnClickListener {
                            order.identifier?.let { identifier ->
                                EventDetailsFragmentDirections.actionEventDetailsToOrderDetail(
                                    eventId = safeArgs.eventId,
                                    orderId = order.id,
                                    orderIdentifier = identifier
                                )
                            }?.let { navigation ->
                                findNavController(rootView).navigate(navigation)
                            }
                        }
                        return@forEach
                    }
                }
```

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -154,6 +156,30 @@ class EventDetailsFragment : Fragment() {
...
+            .nonNull()
+            .observe(viewLifecycleOwner, Observer {
+                it.forEach { order ->
+                    if (order.event?.id == safeArgs.eventId) {
+                        rootView.alreadyRegisteredLayout.visibility = View.VISIBLE
```
anhanh11001: You should also add a variable to retain the visibility state of alreadyRegisteredLayout in the ViewModel.
Just add var isAlreadyOrdered = false  in the viewModel-> change that to true when `order.event?.id == safeArgs.eventId` and set `rootView.alreadyRegisteredLayout.isVisible = viewModel.isAlreadyOrdered` inside onCreateView()

liveHarshit: > `rootView.alreadyRegisteredLayout.isVisible = viewModel.isAlreadyOrdered` inside onCreateView()

Okay, then how it will update when the value is set to true if I write it outside observer?

liveHarshit: If the null check is there for orders from the view model, then this another variable is not required.

anhanh11001: oh yeah that's true

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -105,6 +106,7 @@ class EventDetailsFragment : Fragment() {
     private val LINE_COUNT: Int = 3
     private val LINE_COUNT_ORGANIZER: Int = 2
     private var menuActionBar: Menu? = null
+    private var orderId: Long? = null
```
anhanh11001: this is unused I think. The rest is good

Issue #1948, opened 2019-06-12
Improve UI on SearchFragment and related Search section


anhanh11001:
**Describe the feature you'd like**
Improve UI on SearchFragment and related Search section

**Would you like to work on the issue?**
Yes

PR #1950, opened 2019-06-13
feat: Improve Search Section

anhanh11001:
Details:
- Change layout, text size,...
- Remove Action bar and replace by toolbar for better looks
- Set up better search section in SearchFragment

Fixes: #1948

Screenshots for the change:


anhanh11001:
Please review the code and the UI first, for the background of SearchFragment, I'll make a better one later today and push another commit. The current background is temporary

ShridharGoel:
In my opinion, it would be better if the options under `Filter Search` are on the left, with each option having the checkbox first and then the text, i.e., checkbox on the left of text.

anhanh11001:
Updated:

PR #1977, opened 2019-06-18
fix: Check for device location settings

### LINKED ISSUE #1971: Location progress is continued showing
**Describe the bug**
Location progress is continued showing if the device location is off

yes

liveHarshit:
Fixes #1971

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
@@ -29,6 +29,7 @@ import kotlinx.android.synthetic.main.fragment_search_location.view.toolbar
 import kotlinx.android.synthetic.main.fragment_search_location.view.shimmerSearchEventTypes
 import org.fossasia.openevent.general.R
 import org.fossasia.openevent.general.utils.Utils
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/search/SearchLocationFragment.kt
```diff
@@ -105,6 +111,19 @@ class SearchLocationFragment : Fragment() {
         }
     }
 
+    private fun isLocationEnabled(): Boolean {
```
nikit19: This logic can be re used. Using it from a  util function would be better

app/src/main/java/org/fossasia/openevent/general/utils/Utils.kt
```diff
@@ -204,4 +207,17 @@ object Utils {
             else -> cardType.NONE
         }
     }
+
+    fun checkLocationSettings(context: Context): Boolean {
```
iamareebjamal: `checkLocationSettings` is a verb, hence, this function should return nothing

If you want this to return boolean, rename to `isLocationEnabled`

liveHarshit: Updated

PR #1993, opened 2019-06-22
fix: Snackbar message in profile

### LINKED ISSUE #1979: Snackbar message is displaying incorrectly in profile
**Describe the bug**
Snackbar message is displaying incorrectly in profile. On create is called every time coming back and the previus message is displaying.

yes

liveHarshit:
Fixes #1979


liveHarshit:
This is a workaround.
In view model - 
```java
private val mutableMessage = SingleLiveEvent<String>()
val message: LiveData<String> = mutableMessage
```

In fragment's on create view - 
```java
profileViewModel.message
            .nonNull()
                rootView.snackbar(it)
                profileViewModel.mutableMessage.postValue(null)
            })
```

Any better solution?

Issue #1951, opened 2019-06-13
Add more animation on fragment


anhanh11001:
**Describe the feature you'd like**
- Changing elevation of toolbar after scrolling
- slide up and down when showing/hiding sections

**Additional context**
Should be worked after #1950 got merged

**Would you like to work on the issue?**
Yes

PR #2007, opened 2019-06-25
feat: Add toolbar for fragments + animation

anhanh11001:
Details:
- Add toolbar when scrolling in fragments
- Add animation to show and hide toolbar

Fixes: #1951 

Screenshots for the change:
ezgif-2-668b37a4a56f 
 ezgif-2-977580aedee5

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
@@ -7,6 +7,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
 import androidx.core.view.isVisible
+import androidx.core.widget.NestedScrollView
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.recyclerview.widget.GridLayoutManager
@@ -24,6 +25,10 @@ import kotlinx.android.synthetic.main.fragment_events.view.scrollView
 import kotlinx.android.synthetic.main.fragment_events.view.notification
 import kotlinx.android.synthetic.main.fragment_events.view.swiperefresh
 import kotlinx.android.synthetic.main.fragment_events.view.newNotificationDot
+import kotlinx.android.synthetic.main.fragment_events.view.toolbar
+import kotlinx.android.synthetic.main.fragment_events.view.toolbarLayout
+import kotlinx.android.synthetic.main.fragment_events.view.newNotificationDotToolbar
+import kotlinx.android.synthetic.main.fragment_events.view.notificationToolbar
 import org.fossasia.openevent.general.R
 import org.fossasia.openevent.general.BottomIconDoubleClick
 import org.fossasia.openevent.general.common.EventClickListener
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -184,6 +196,40 @@ class EventsFragment : Fragment(), BottomIconDoubleClick {
             onFavFabClick = favFabClickListener
             onHashtagClick = hashTagClickListener
         }
+
+        rootView.scrollView.setOnScrollChangeListener { _: NestedScrollView?, _: Int, scrollY: Int, _: Int, _: Int ->
+            if (scrollY > rootView.locationTextView.y + rootView.locationTextView.height) {
+                if (!rootView.toolbarLayout.isVisible) showToolbar()
```
nikit19: these 2 if statements can be combined

app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -184,6 +196,40 @@ class EventsFragment : Fragment(), BottomIconDoubleClick {
...
+            } else {
+                if (rootView.toolbarLayout.isVisible) hideToolbar()
```
nikit19: `else if`

app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
anhanh11001: fixed

app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
anhanh11001: fixed

app/src/main/java/org/fossasia/openevent/general/event/EventsFragment.kt
```diff
@@ -157,6 +166,9 @@ class EventsFragment : Fragment(), BottomIconDoubleClick {
         rootView.notification.setOnClickListener {
             findNavController(rootView).navigate(EventsFragmentDirections.actionEventsToNotification())
         }
+        rootView.notificationToolbar.setOnClickListener {
+            findNavController(rootView).navigate(EventsFragmentDirections.actionEventsToNotification())
```
liveHarshit: Duplication can be removed

anhanh11001: notification is an ImageView shown when toolbar is hide, notificationToolbar is an icon in the toolbar when toolbar is shown, so it's not duplicate

liveHarshit: oh, that's not what I mean. I'm talking about `findNavController(rootView).navigate(EventsFragmentDirections.actionEventsToNotification())`. But it is okay. :+1:

anhanh11001: oh okay, I'm fixing now, besides, build is failing so I need to make some fixes as well

PR #2008, opened 2019-06-26
feat: Implement location feature in f-droid build

### LINKED ISSUE #2006: Add current location feature for f froid build

yes

liveHarshit:
Fixes #2006 
 - Remove flavor specific module as not required.


iamareebjamal:
> Remove flavor specific module as not required.

Why not required?

liveHarshit:
> > Remove flavor specific module as not required.
> 
> Why not required?

There was only one module which is configured for location feature, now the location feature is implemented in both variant in the same way, so I have removed the blank module

iamareebjamal:
How was it implemented previously in Google Play Variant?

liveHarshit:
> How was it implemented previously in Google Play Variant?

There was a ViewModel: WelcomeViewMode same copy of GeoLocationViewModel in flavor specific view model. Now I use the GeoLocationViewModel in place of that.


> now the location feature is implemented in both variant in the same way

The same way means ViewModel type is same but location service implementation is different. No change in location service implementation for play variant, I have just implemented the location service class for fdroid variant.

liveHarshit:
Also, there were two welcome fragments for different variant, which is really not required now.

app/src/main/AndroidManifest.xml
```diff
@@ -7,6 +7,8 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
```
iamareebjamal: No use of fine AFAIK, or is there?

liveHarshit: Here `LocationManager.NETWORK_PROVIDER` is used so fine location is required.

liveHarshit: According to https://developer.android.com/guide/topics/location/strategies.html

PR #2012, opened 2019-06-26
feat: Add pagination

### LINKED ISSUE #84: Implement pagination in the app

anhanh11001:
Details:
- Add pagination for EventsFragment
- Remove unused methods and variables
- Fix layout to match pagination

Fixes #84
Fixes #1681 

Screenshots for the change:
ezgif-4-191763ff2c27
ezgif-4-f6209224c047

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.general.di
 
+import androidx.paging.PagedList
 import androidx.room.Room
 import com.facebook.stetho.okhttp3.StethoInterceptor
 import com.fasterxml.jackson.databind.DeserializationFeature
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
```diff
@@ -250,6 +251,16 @@ val networkModule = module {
         objectMapper
     }
 
+    single {
+        PagedList
+            .Config
+            .Builder()
+            .setPageSize(5)
```
nikit19: so we are requesting for more data after 5 elements right? I guess this is too less. We can load 15-20 events at time if it doesn't take too much time to load.

iamareebjamal: I think it is fine. Much better for mobile data. Will test locally

nikit19: @iamareebjamal did you test? This PR looks good otherwise, let's finalise it

PR #2019, opened 2019-06-28
chore(deps): Bump logging-interceptor to 4.0.0 and fix errors

liveHarshit:
Fixes #2018

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.github.jasminb.jsonapi.ResourceConverter
 import com.github.jasminb.jsonapi.retrofit.JSONAPIConverterFactory
-import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import org.fossasia.openevent.general.BuildConfig
@@ -250,8 +249,6 @@ val networkModule = module {
         objectMapper
     }
 
-    single { RequestAuthenticator(get()) as Interceptor }
-
     single {
         val connectTimeout = 15 // 15s
         val readTimeout = 15 // 15s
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
```diff
@@ -259,7 +259,7 @@ val networkModule = module {
         val builder = OkHttpClient().newBuilder()
             .connectTimeout(connectTimeout.toLong(), TimeUnit.SECONDS)
             .readTimeout(readTimeout.toLong(), TimeUnit.SECONDS)
-            .addInterceptor(get())
+            .addInterceptor(RequestAuthenticator(get()) as Interceptor)
```
iamareebjamal: Casting is bad

liveHarshit: Sorry, It is not required now as the response is not nullable now. Updated.

app/src/main/java/org/fossasia/openevent/general/di/Modules.kt
```diff
@@ -259,7 +258,7 @@ val networkModule = module {
         val builder = OkHttpClient().newBuilder()
             .connectTimeout(connectTimeout.toLong(), TimeUnit.SECONDS)
             .readTimeout(readTimeout.toLong(), TimeUnit.SECONDS)
-            .addInterceptor(get())
+            .addInterceptor(RequestAuthenticator(get()))
```
iamareebjamal: Why not just get()?

liveHarshit: Displaying error-
```
@JvmName public final inline fun addInterceptor(crossinline block: (chain: Interceptor.Chain) -> Response): OkHttpClient.Builder defined in okhttp3.OkHttpClient.Builder
public final fun addInterceptor(interceptor: Interceptor): OkHttpClient.Builder defined in okhttp3.OkHttpClient.Builder
```

iamareebjamal: You are passing `RequestAuthenticator(get())` and `single { RequestAuthenticator(get()) }` is already defined above.  Either remove that or use get

liveHarshit: Updated

Issue #2026, opened 2019-06-29
wrong order status when holding ticket for user


anhanh11001:
**Describe the bug**
Currently, with time counter on AttendeeFragment, "pending" order is created, which is not correct as from the implementation of the frontend, "initializing" order status should be used

**Would you like to work on the issue?**
Yes

PR #2032, opened 2019-06-30
fix: wrong order status

anhanh11001:
Details:
- When initializing an order, the order status used is "INITIALIZING", not "PENDING".
- Don't have to cancel order when time counter goes down as the server automatically hanndle it

Fixes: #2026


iamareebjamal:
Also fix the timer minutes, it's been changed on the server

anhanh11001:
What change is needed to be made? 

From what I understand, it is already handled correctly as the time counts down based on the registration time entered by the organizer, and when time goes up, the app redirects the user to the previous fragment. 

The server turn "initializing" order into "expired" whenever a GET request for order it made and order start time + registration time < current time

iamareebjamal:
The countdown time has changed to 15 minutes or so. Confirm on gitter channel

PR #2039, opened 2019-07-01
fix: Back button in location fragment

### LINKED ISSUE #2011: Back button is not working in search location fragment

yes

liveHarshit:
Fixes #2011


liveHarshit:
It was due to the text listener, search view or edit text were always in focus.

Issue #2041, opened 2019-07-01
App crashes in TicketsFragment - no internet


anhanh11001:
**Describe the bug**
App crashes in TicketsFragment when there is no internet

**Screenshots**

**Would you like to work on the issue?**
Yes

PR #2042, opened 2019-07-02
fix: app crashes in TicketsFragment

anhanh11001:
Details:
- Hide discount code when there is no internet in TicketFragment
- Fix extra in querystring orderApi
- Fix crashes due to wrong way to handle retaining ticket number in TicketAdapter

Fixes: #2041


app/src/main/java/org/fossasia/openevent/general/ticket/TicketsViewModel.kt
```diff
@@ -43,6 +43,7 @@ class TicketsViewModel(
     private val mutableTicketTableVisibility = MutableLiveData<Boolean>()
     val ticketTableVisibility: LiveData<Boolean> = mutableTicketTableVisibility
     val ticketIdAndQty = MutableLiveData<List<Triple<Int, Int, Float>>>()
+    var discountCodeCurrentLayout = APPLY_DISCOUNT_CODE
```
iamareebjamal: Why is this var?

anhanh11001: to retain current discount code layout on screen rotation or coming back from AttendeeFragment

PR #2065, opened 2019-07-04
fix: Malformed name of aab file

### LINKED ISSUE #2064: Malformed name of aab file
aab file has wrong name - eventyay-attendee-dev-aab

liveHarshit:
Fixes #2064

### DIFF
--- a/scripts/update-apk.sh
+++ b/scripts/update-apk.sh
@@ -18,7 +18,7 @@ git clone --quiet --branch=apk https://fossasia:$GITHUB_API_KEY@github.com/fossa
 cd apk
 
 if [ "$TRAVIS_BRANCH" == "$PUBLISH_BRANCH" ]; then
-	/bin/rm -f  open-event-attendee-master-*
+	/bin/rm -f  *
 else
 	/bin/rm -f open-event-attendee-dev-*
 fi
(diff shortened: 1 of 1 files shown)


scripts/update-apk.sh
```diff
@@ -45,6 +45,12 @@ if [ "$TRAVIS_BRANCH" == "$DEPLOY_BRANCH" ]; then
 	done
 fi
 
+if [ "$TRAVIS_BRANCH" == "$PUBLISH_BRANCH" ]; then
+	mv eventyay-attendee-master-aab eventyay-attendee-master.aab
+else
+	mv eventyay-attendee-dev-aab eventyay-attendee-dev.aab
+fi
```
iamareebjamal: This is not the right solution

iamareebjamal:
Also, on merge of master, every file should be deleted or else, the old named files will never be deleted

liveHarshit:
> Also, on merge of master, every file should be deleted

So it should be - 
```bash
if [ "$TRAVIS_BRANCH" == "$PUBLISH_BRANCH" ]; then
	/bin/rm -f  *
else
	/bin/rm -f open-event-attendee-dev-*
fi
```

iamareebjamal:
Yes

liveHarshit:
> Yes

Updated

Issue #2028, opened 2019-06-29
Add sorting/filtering order/tickets based order status


anhanh11001:
**Describe the feature you'd like**
Add sorting/filtering order/tickets based order status

**Additional context**
Should be worked after #2026 is finished

**Would you like to work on the issue?**
Yes

PR #2069, opened 2019-07-05
feat: Add filtering and sorting dialog

anhanh11001:
Details:
- Add filtering and sorting orders

Fixes: #2028

Screenshots for the change:
ezgif-4-671bb21821f7
ezgif-4-cdee3c6ea26f

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/order/ExpiredOrderFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/order/ExpiredOrderFragment.kt
@@ -2,20 +2,26 @@ package org.fossasia.openevent.general.order
 
 import android.os.Bundle
 import android.view.LayoutInflater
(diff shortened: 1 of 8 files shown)


anhanh11001:
> Can't be `ExpiredOrderFragment.kt` and `OrderUnderUserFragment` merge?

What do you mean by merged? If it is about having the same Fragment than I'm not sure because they have different layouts

liveHarshit:
> > Can't be `ExpiredOrderFragment.kt` and `OrderUnderUserFragment` merge?
> 
> What do you mean by merged? If it is about having the same Fragment than I'm not sure because they have different layouts

Only the toolbar layout is different which can be hidden. But okay it will be complex, needs to handle navigation bar also.

iamareebjamal:
Open an issue to make the label small as it will look truncated on almost all devices

PR #2076, opened 2019-07-06
feat: Integrate in app update API

### LINKED ISSUE #1823: Integrate In-App Update API
References
https://developer.android.com/guide/app-bundle/in-app-updates

- [x] Yes

liveHarshit:
Fixes #1823

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
@@ -7,6 +7,9 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.navigation.NavController
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.NavigationUI.setupWithNavController
+import com.google.android.play.core.appupdate.AppUpdateManagerFactory
+import com.google.android.play.core.install.model.AppUpdateType
+import com.google.android.play.core.install.model.UpdateAvailability
 import kotlinx.android.synthetic.main.activity_main.navigation
 import kotlinx.android.synthetic.main.activity_main.mainFragmentCoordinatorLayout
 import org.fossasia.openevent.general.auth.RC_CREDENTIALS_READ
@@ -23,6 +26,7 @@ const val EVENT_IDENTIFIER = "eventIdentifier"
 const val VERIFICATION_TOKEN = "verificationToken"
 private const val VERIFY = "verify"
 private const val TOKEN = "token"
+private const val UPDATE_REQUEST_CODE = 3
 
 class MainActivity : AppCompatActivity() {
     private lateinit var navController: NavController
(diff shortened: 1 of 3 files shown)


liveHarshit:
@iamareebjamal We need a mock update with the same signing key to debug the implementation.

iamareebjamal:
The private key is with Google since we use app signing. Search a workaround of what to do in that case

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -44,6 +48,22 @@ class MainActivity : AppCompatActivity() {
...
+    }
+
+    private fun checkForUpdates() {
+        val appUpdateManager = AppUpdateManagerFactory.create(this)
+        val appUpdateInfoTask = appUpdateManager.appUpdateInfo
+        appUpdateInfoTask.addOnSuccessListener { appUpdateInfo ->
+            if (appUpdateInfo.updateAvailability() == UpdateAvailability.UPDATE_AVAILABLE) {
+                appUpdateManager.startUpdateFlowForResult(
+                    appUpdateInfo,
+                    AppUpdateType.IMMEDIATE,
```
nikit19: This should be a `FLEXIBLE` update instead. `IMMEDIATE` updates are done only when there are some major crashes in the app.

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
```diff
@@ -127,6 +147,15 @@ class MainActivity : AppCompatActivity() {
         } else {
             super.onActivityResult(requestCode, resultCode, data)
         }
+        if (requestCode == UPDATE_REQUEST_CODE) {
+            if (resultCode != Activity.RESULT_OK) {
+                val dialog = androidx.appcompat.app.AlertDialog.Builder(this)
```
nikit19: Why? Rather call the `checkForUpdates` methods again.  This dialog doesn't even update the app

iamareebjamal:
> You can test the app for in app updates by creating a release apk with the same keystore as in production.

We don't have production keystore.

The private key is with Google since we use app signing. Search a workaround of what to do in that case

app/src/main/java/org/fossasia/openevent/general/MainActivity.kt
liveHarshit: Why call check for update method again when a user cancel the update? It will just alert the user why the update is important after canceling it. We can add a button which will redirect to play store to update after in-app update is canceled.

nikit19: I wrote the above comment assuming that we were doing an `immediate` update. If that had been the case we want to force the user to update the app  (eg: there is a security vulnerability in the app and the user is trying to exploit it)
But since this is not the case, for  a flexible update you can call the `checkForUpdates` methods if the update fails.

`if (resultCode == ActivityResult.RESULT_IN_APP_UPDATE_FAILED) {
                checkForUpdates()
            }`

iamareebjamal:
What'd be the point of changing the key though?

nikit19:
If we want to test this feature that's the only way to do it.
Although this doesn't look like a high priority issue.

Issue #2085, opened 2019-07-09
Unable to click back press in SearchLocationFragment


anhanh11001:
**Describe the bug**
Unable to click back press in SearchLocationFragment

**Would you like to work on the issue?**
Yes

anhanh11001:
@liveHarshit Can you check if it is reproducible in your device, it is in my and it's weird cause I thought you sent a PR earlier

liveHarshit:
Yes, It was solved at that time. Will check again :+1:

PR #2096, opened 2019-07-10
fix: SearchLocationFragment back press not working

anhanh11001:
Details:
- Use EditText instead of SearchView

Tested in API 22, 26,28

Fixes: #2085


liveHarshit:
So currently it is working fine with Nougat 7.1 but not with API Q. IMO there were two reasons for the back press is not working.

PR #2104, opened 2019-07-11
feat: Fetch settings and divide events view model

### LINKED ISSUE #2083: Break events view model into smaller ones

yes

### LINKED ISSUE #2101: Fetch timer from settings
- Timer for ticket registration should be according to the organizer preference.
- Message for the timer is displaying wrong. 

<!-- Please let us know if you can work on it or the issue should be assigned to someone else. --yes

liveHarshit:
Fixes #2101 
Fixes #2083 

<p align="center">
 
</p>

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
@@ -117,8 +117,6 @@ import java.util.Calendar
 import java.util.Currency
 import kotlin.collections.ArrayList
 
-private const val COUNT_DOWN_TIME = 15 // in minutes
-
 class AttendeeFragment : Fragment(), ComplexBackPressFragment {
 
     private lateinit var rootView: View
@@ -260,9 +258,13 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
             .observe(viewLifecycleOwner, Observer {
                 loadEventDetailsUI(it)
                 setupPaymentOptions(it)
-                if (attendeeViewModel.pendingOrder.value != null) {
-                    setupCountDownTimer(it)
-                }
+            })
+
+        attendeeViewModel.getSettings()
+        attendeeViewModel.orderExpiryTime
+            .nonNull()
+            .observe(viewLifecycleOwner, Observer {
+                setupCountDownTimer(it)
             })
 
         val currentEvent = attendeeViewModel.event.value
@@ -271,33 +273,22 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
         else {
             setupPaymentOptions(currentEvent)
             loadEventDetailsUI(currentEvent)
-            if (attendeeViewModel.pendingOrder.value != null) {
-                setupCountDownTimer(currentEvent)
-            }
         }
     }
 
     private fun setupPendingOrder() {
-        attendeeViewModel.pendingOrder
-            .nonNull()
-            .observe(viewLifecycleOwner, Observer {
-                attendeeViewModel.event.value?.let {
-                    setupCountDownTimer(it)
-                }
-            })
-
         val currentPendingOrder = attendeeViewModel.pendingOrder.value
         if (currentPendingOrder == null) {
             attendeeViewModel.initializeOrder(safeArgs.eventId)
         }
     }
 
-    private fun setupCountDownTimer(event: Event) {
+    private fun setupCountDownTimer(orderExpiryTime: Int) {
         rootView.timeoutCounterLayout.visibility = View.VISIBLE
         rootView.timeoutInfoTextView.text =
-            getString(R.string.ticket_timeout_info_message, event.orderExpiryTime.toString())
+            getString(R.string.ticket_timeout_info_message, orderExpiryTime.toString())
 
-        val timeLeft: Long = if (attendeeViewModel.timeout == -1L) COUNT_DOWN_TIME * 60 * 1000L
+        val timeLeft: Long = if (attendeeViewModel.timeout == -1L) orderExpiryTime * 60 * 1000L
                                 else attendeeViewModel.timeout
         timer = object : CountDownTimer(timeLeft, 1000) {
             override fun onFinish() {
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
```diff
@@ -297,7 +295,7 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
         rootView.timeoutInfoTextView.text =
             getString(R.string.ticket_timeout_info_message, event.orderExpiryTime.toString())
 
-        val timeLeft: Long = if (attendeeViewModel.timeout == -1L) COUNT_DOWN_TIME * 60 * 1000L
+        val timeLeft: Long = if (attendeeViewModel.timeout == -1L) event.orderExpiryTime * 60 * 1000L
```
iamareebjamal: This was moved to settings, so it won't be available in event

liveHarshit: But how it is working currently. I've tested it, frontend is also using the same endpoint.

iamareebjamal: Working on dev or prod?

iamareebjamal: https://github.com/fossasia/open-event-server/pull/6083

liveHarshit: > Working on dev or prod?

production, while creating an event, time is asking form the organizer.

iamareebjamal: Yeah, so again, this was moved to settings, so it won't be available in event. (10 days ago)

Always consider dev to be latest. It'll break when server is updated.

That's why it's asked for client devs to only use dev instance. You must be ready for any breaking change being applied on server. If you'll use prod environment, how will you deal with new changes?

liveHarshit: > Always consider dev to be latest. It'll break when server is updated.

Okay, I should also check with the dev instance. It was during creating the meetup event. Will update.

liveHarshit: Only admin can reterive order expiry time - 
image

iamareebjamal: Open an issue to allow clients to see order expiry time. If this is true, then I guess timer on frontend is wrong as well

liveHarshit: Will fetch setting and test +1
Should I fetch setting at the starting of the app and store into the database or fetch every time when required?

liveHarshit:
https://github.com/fossasia/open-event-server/issues/6185

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
iamareebjamal: Save into DB and when retrieving, first load from DB and then send a request to refresh

liveHarshit: > Save into DB and when retrieving, first load from DB and then send a request to refresh

One problem will there if I load time from the database first and after that refresh it from the server then count down timer will start again by observing new value.

liveHarshit: Currently, I'm directly fetching settings at the time of ordering tickets. Updated.

iamareebjamal: You will not show it again, it'll be shown next time user opens the order screen. If even you are directly fetching, save in DB

liveHarshit:
Updated

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
```diff
@@ -260,9 +258,13 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
             .observe(viewLifecycleOwner, Observer {
                 loadEventDetailsUI(it)
                 setupPaymentOptions(it)
-                if (attendeeViewModel.pendingOrder.value != null) {
-                    setupCountDownTimer(it)
-                }
+            })
+
+        attendeeViewModel.getSettings()
```
iamareebjamal: There is no guarantee that settings have been fetched till the next statement is executed

liveHarshit: Then onErrorResumeNext will execute and settings will fetch from api.

iamareebjamal: Still there is no guarantee that settings have been fetched till the next statement is executed, as getsettings is asynchronous

liveHarshit: Sorry, I don't understand which next statement. During fetching/getting settings progress dialog is displaying and if settings fetch successfully or there is an error(=15), value for mutableOrderExpiryTime is set and only then countdown timer will start.

iamareebjamal: The statement which is immediately proceeding getSettings

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -391,6 +395,21 @@ class AttendeeViewModel(
...
+        compositeDisposable += settingsService.getSettings()
+            .withDefaultSchedulers()
+            .doOnSubscribe {
+                mutableProgress.value = true
+            }.doFinally {
+                mutableProgress.value = false
+            }.subscribe({
+                mutableOrderExpiryTime.value = it.orderExpiryTime
+            }, {
+                mutableOrderExpiryTime.value = 15
```
anhanh11001: use const value here

liveHarshit:
Updated

iamareebjamal:
Approving but there is a need of polishing

Issue #2084, opened 2019-07-09
Add paging in OrdersUnderUserFragment


anhanh11001:
**Describe the feature you'd like**
Add paging in OrdersUnderUserFragment

**Would you like to work on the issue?**
Yes

PR #2106, opened 2019-07-11
feat: Add paging for order

anhanh11001:
Details:
- Add paging for Order
- Use chaining RxJava for getting orders and events
- Change UI of OrdersUnderUserFragment and ExpiredOrdersFragment
- Remove unused methods

Fixes: #2084 

Screenshots for the change:
ezgif-2-5a3f63cf39bf
ezgif-2-e50514a06128


app/src/main/java/org/fossasia/openevent/general/order/OrdersUnderUserViewModel.kt
```diff
@@ -3,112 +3,103 @@ package org.fossasia.openevent.general.order
...
-                        EventUtils.getTimeInMilliSeconds(it.first.endsAt, null) < System.currentTimeMillis() }
+                    currentPagedOrdersAndEvents.addAll(it)
+                    mutableEventAndOrderPaged.value = currentPagedOrdersAndEvents
                 }
-                if (finalList.isEmpty()) mutableNoTickets.value = true
-                mutableEventAndOrder.value = finalList
             }, {
-                mutableShowShimmerResults.value = false
                 mutableMessage.value = resource.getString(R.string.list_events_fail_message)
                 Timber.d(it, "Failed  to list events under a user ")
```
liveHarshit: I think it will never through the error, the error should handle in the order data source class.

anhanh11001: make sense, I'll move it to the data source class

anhanh11001: updated

PR #2110, opened 2019-07-12
fix: Display unauthorization error while loading events

### LINKED ISSUE #2109: Progress continues on unauthorization error
**Describe the bug**
Progress continues on authorization error while loading events,

```
E/EventsDataSource$createObservable: Fail on fetching page of events
    retrofit2.adapter.rxjava2.HttpException: HTTP 401 UNAUTHORIZED
```

yes

### LINKED ISSUE #2111: App crashes due to empty intent

yes

liveHarshit:
Fixes #2109 
Fixes #2111

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
@@ -22,7 +22,9 @@ import org.fossasia.openevent.general.data.Resource
 import org.fossasia.openevent.general.event.paging.EventsDataSourceFactory
 import org.fossasia.openevent.general.notification.NotificationService
 import org.fossasia.openevent.general.search.location.SAVED_LOCATION
+import org.fossasia.openevent.general.utils.HttpErrors
 import org.fossasia.openevent.general.utils.extensions.withDefaultSchedulers
+import retrofit2.HttpException
 import timber.log.Timber
 
 const val NEW_NOTIFICATIONS = "newNotifications"
@@ -134,7 +136,25 @@ class EventsViewModel(
                     }
                 }
             }, {
+                if (it is HttpException) {
+                    if (authHolder.isLoggedIn() && it.code() == HttpErrors.UNAUTHORIZED) {
+                        logoutAndRefresh()
+                    }
+                }
                 Timber.e(it, "Error fetching notifications")
+            })
+    }
+
+    private fun logoutAndRefresh() {
+        compositeDisposable += authService.logout()
+            .withDefaultSchedulers()
+            .subscribe({
+                loadLocationEvents()
+                syncNotifications()
+            }, {
+                mutableProgress.value = false
+                Timber.e(it, "Error while logout")
+                mutableMessage.value = resource.getString(R.string.error)
             })
     }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventsViewModel.kt
```diff
@@ -126,6 +127,10 @@ class EventsViewModel(
                     }
                 }
             }, {
+                if ((it as HttpException).code() == 401) {
```
ShridharGoel: Define a constant

ShridharGoel: Define constant in HttpErrors.kt

ShridharGoel: Define constant in `HttpErrors.kt`.

iamareebjamal: As said by codacy, this is an Unsafe cast. You can't know if the exception is HttpException, it can be IOException IllegalArgumentException, anything and you just casting it is waiting for a crash to happen.

Also, it's the developer's responsibility to handle any error except 500, where you can show Unfortunately, something went wrong, because you actually don't know what happened. 

All other errors must be handled. As a user, what am I supposed to do with a toast that says `HTTP 401 UNAUTHORIZED`?

I should either see `You must log in to see this {item}`, or if I am already logged in, then `Your session has expired, please login again`

And maybe being redirected to login screen

liveHarshit: > I should either see `You must log in to see this {item}`, or if I am already logged in, then `Your session has expired, please login again`

I was not sure it is because of the user's login state. It shows when I try to change API and do not clear the app data. And events should load if the user is not logged in. I think it is due to the wrong Authorization which needs to re-request the Authenticator.

liveHarshit: > As said by codacy, this is an Unsafe cast. You can't know if the exception is HttpException, it can be IOException IllegalArgumentException, anything and you just casting it is waiting for a crash to happen.

Sorry about that.

iamareebjamal: > I try to change API and do not clear the app data.

Then clear the data when API change happens

liveHarshit: Mean clear the app data programmatically?

liveHarshit: Can we re-request the authenticator and reload events?

iamareebjamal: No, because access token from one API won't work in another. This is very easily handled in Orga App. We simply show login screen anytime we encounter 401. Reason may be expired token or changing API base URL. Fixes every case

liveHarshit: Yes, It shows in case of login state only. For that error, we can log out the user and reload events. Because events should also load in not logged in state. :+1:

liveHarshit: Will fetch setting and test :+1:
Should I fetch setting at the starting of the app and store into the database or fetch every time when required?

iamareebjamal: Why fetching settings?

liveHarshit: Sorry wrong PR.

liveHarshit: Updated

Issue #2114, opened 2019-07-13
Add delete account feature


anhanh11001:
**Describe the feature you'd like**
Add delete account feature

**Would you like to work on the issue?**
Yes

PR #2115, opened 2019-07-13
feat: Add delete account feature

anhanh11001:
Fixes: #2114

Screenshots for the change:
ezgif-2-7c3e3618736e

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragment.kt
@@ -20,6 +20,14 @@ import kotlinx.android.synthetic.main.dialog_change_password.view.newPassword
 import kotlinx.android.synthetic.main.dialog_change_password.view.confirmNewPassword
 import kotlinx.android.synthetic.main.dialog_change_password.view.textInputLayoutConfirmNewPassword
 import kotlinx.android.synthetic.main.dialog_change_password.view.textInputLayoutNewPassword
+import kotlinx.android.synthetic.main.dialog_confirm_delete_account.view.password
+import kotlinx.android.synthetic.main.dialog_delete_account.view.confirmEmailButton
+import kotlinx.android.synthetic.main.dialog_delete_account.view.confirmEmailLayout
+import kotlinx.android.synthetic.main.dialog_delete_account.view.confirmDeleteCheckbox
+import kotlinx.android.synthetic.main.dialog_delete_account.view.confirmDeleteLayout
+import kotlinx.android.synthetic.main.dialog_delete_account.view.deleteButton
+import kotlinx.android.synthetic.main.dialog_delete_account.view.cancelButton
+import kotlinx.android.synthetic.main.dialog_delete_account.view.email
 import kotlinx.android.synthetic.main.fragment_profile.view.login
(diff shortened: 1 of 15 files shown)


app/src/main/res/layout/dialog_confirm_delete_account.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    <com.google.android.material.textfield.TextInputLayout
+        style="@style/Widget.MaterialComponents.TextInputLayout.OutlinedBox.Dense"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:hint="@string/enter_account_password_message"
+        android:layout_marginBottom="@dimen/details_margin_small">
+
+        <com.google.android.material.textfield.TextInputEditText
+            android:id="@+id/password"
```
ShridharGoel: Add `android:inputType="textPassword"`

anhanh11001: updated

app/src/main/java/org/fossasia/openevent/general/auth/ProfileViewModel.kt
```diff
@@ -38,6 +45,43 @@ class ProfileViewModel(private val authService: AuthService, private val resourc
...
+        val query = """[{
+                |   'and':[{
+                |       'name':'is-event-online',
+                |       'op':'eq',
+                |       'val': 'true'
+                |    }]
+                |}]""".trimMargin().replace("'", "\"")
+
+        compositeDisposable += eventService.getEventsOfUser(authService.getId(), query)
+            .withDefaultSchedulers()
```
iamareebjamal: This is not your job. This is server's job

anhanh11001: I thought frontend also did the same. There is no endpoint to check whether an account is deletable or not, so we have to check whether there are active events attached to the account from client-side or not

anhanh11001: So should I just proceed to delete and then when server return delete fail, I'll display the message that there are events attached so the account is not deletable?

iamareebjamal: yes

app/src/main/java/org/fossasia/openevent/general/auth/ProfileFragment.kt
```diff
@@ -182,6 +204,92 @@ class ProfileFragment : Fragment(), BottomIconDoubleClick {
...
+    private fun confirmDeleteAccount() {
+        val progressDialog = progressDialog(context, getString(R.string.deleting_account))
+
+        loginViewModel.progress
+            .nonNull()
+            .observe(viewLifecycleOwner, Observer {
+                progressDialog.show(it)
+            })
+
+        loginViewModel.validPassword
```
liveHarshit: Can you use chaining here, after password validation, directly use the data?

anhanh11001: I don't think so, chaining is more about using result from the previous request for the second request, should not be nested

anhanh11001:
Reopening as mistakenly closed

Issue #2126, opened 2019-07-15
Illogical information displayed in Ticket and Event section


anhanh11001:
**Describe the bug**
- Gender picker should be turned into a spinner instead of edit text
- Empty white unused space about event detail of order detail section
- Empty organizer name in Tickets Fragment
- Confusing Cancel/Continue Order
- Subtotal/Total ticket price in AttendeeFragment
- Display button "Register", "Pay now"

**Would you like to work on the issue?**
Yes

PR #2129, opened 2019-07-15
fix: Illogical information displayed in Ticket and Event section

anhanh11001:
Details:
- Gender picker should be turned into a spinner instead of edit text
- Empty white unused space about event detail of order detail section
- Empty organizer name in Tickets Fragment
- Confusing Cancel/Continue Order
Subtotal/Total ticket price in AttendeeFragment
- Display button "Register", "Pay now"
- Fix filter for custom form

Fixes: #2126 

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewHolder.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewHolder.kt
@@ -5,6 +5,9 @@ import android.text.Editable
 import android.text.InputType
 import android.text.SpannableStringBuilder
 import android.text.TextWatcher
+import android.view.View
+import android.widget.AdapterView
+import android.widget.ArrayAdapter
 import androidx.core.view.isVisible
 import com.google.android.material.textfield.TextInputEditText
 import com.google.android.material.textfield.TextInputLayout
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewHolder.kt
```diff
@@ -135,10 +117,17 @@ class AttendeeViewHolder(private val binding: ItemAttendeeBinding) : RecyclerVie
...
         }
     }
-    private fun setupCustomFormWithFields(form: CustomForm, textWatcher: TextWatcher) {
+    private fun setupCustomFormWithFields(
+        form: CustomForm,
+        textWatcher: TextWatcher,
+        attendee: Attendee,
+        ticket: Ticket,
+        eventId: Long,
+        position: Int
```
iamareebjamal: These are a lot of parameters, why weren't they required previously, why are they required now?

anhanh11001: Previously, it is only about setting up EditText view, for now, it is Spinner so I need to add a few more parameters just to set up listener, save gender option on screen rotation,... Let me see if I can better this while resolving conflicts

anhanh11001:
> Although changes you have made improve the codebase and it is good practice, but they should be made as a separate PR so that they don't pollute git blame and each commit solves a specific issue
> 
> Please keep in mind from now on

Yes I will keep that in mind

PR #2136, opened 2019-07-17
fix: Make billing information compulsory for paid orders

### LINKED ISSUE #2134: Make billing info compulsory for paid orders

liveHarshit:
Fixes #2134

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
@@ -59,8 +59,6 @@ import kotlinx.android.synthetic.main.fragment_attendee.view.cardNumber
 import kotlinx.android.synthetic.main.fragment_attendee.view.acceptCheckbox
 import kotlinx.android.synthetic.main.fragment_attendee.view.countryPicker
 import kotlinx.android.synthetic.main.fragment_attendee.view.billingInfoContainer
-import kotlinx.android.synthetic.main.fragment_attendee.view.billingInfoCheckboxSection
-import kotlinx.android.synthetic.main.fragment_attendee.view.billingEnabledCheckbox
 import kotlinx.android.synthetic.main.fragment_attendee.view.billingCity
 import kotlinx.android.synthetic.main.fragment_attendee.view.billingCompany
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
```diff
@@ -691,7 +684,7 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
             rootView.email.checkEmpty()
 
         var checkBillingInfo = true
-        if (rootView.billingEnabledCheckbox.isChecked) {
+        if (safeArgs.amount > 0) {
```
iamareebjamal: Invalid check. Please check if all tickets are free or donation. Because amount can be 0 due to application of discount code as well

liveHarshit: So no billing info for donation tickets?

iamareebjamal: Sorry, there will be 2 checks, if all are free/donation and amount == 0, then no billing info

liveHarshit: Updated

app/src/main/java/org/fossasia/openevent/general/ticket/TicketsFragment.kt
```diff
@@ -220,15 +220,16 @@ class TicketsFragment : Fragment() {
...
         ticketsViewModel.mutableAmount.value = null
         findNavController(rootView).navigate(TicketsFragmentDirections.actionTicketsToAttendee(
             eventId = safeArgs.eventId,
             ticketIdAndQty = wrappedTicketAndQty,
             currency = safeArgs.currency,
-            amount = totalAmount
+            amount = totalAmount,
+            hasPaidTickets = ticketsViewModel.hasPaidTickets
         ))
+        ticketsViewModel.hasPaidTickets = false
```
iamareebjamal: Why?

liveHarshit: If we go back and change the ticket type, then it will remain true. Another solution is to set false in on create.

PR #2144, opened 2019-07-18
chore: Bump nav_version and set jvm target to 1.8

liveHarshit:
Fixes #2141 

Reference issue: https://issuetracker.google.com/issues/137782949
Solution: https://issuetracker.google.com/issues/137268935

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -94,12 +94,16 @@ android {
     }
     lintOptions {
         disable 'MissingTranslation'
+        warning 'InvalidPackage'
     }
     androidExtensions {
         experimental = true
     }
     aaptOptions {
         cruncherEnabled = false
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
     }
 }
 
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -94,13 +94,17 @@ android {
     }
     lintOptions {
         disable 'MissingTranslation'
+        disable 'InvalidPackage'
```
ShridharGoel: Why?

ShridharGoel: If this cannot be avoided then consider using `warning 'InvalidPackage'`
https://stackoverflow.com/questions/21827004/gradle-dagger-lint-ignore-by-package#comment58243384_21985399

liveHarshit: Updated. Another solution is: https://github.com/getsentry/sentry-java/issues/571#issuecomment-378295015

iamareebjamal:
Test app on API < 24

liveHarshit:
> Test app on API < 24

Tested in API 23, working fine.

Issue #2182, opened 2019-07-25
Add sliding up/down animation on showing View


anhanh11001:
**Describe the feature you'd like**
Add sliding up/down animation on showing View

**Would you like to work on the issue?**
Yes

PR #2186, opened 2019-07-26
feat: Add slide up down animation

anhanh11001:
Details:
- Add slide up/down animation for View,TextView

Fixes: #2182

Screenshots for the change:
ezgif-2-2b2a603bbce3
ezgif-2-f1470bfb072f

### DIFF
--- a/app/src/main/res/layout/fragment_speakers_call_proposal.xml
+++ b/app/src/main/res/layout/fragment_speakers_call_proposal.xml
@@ -45,163 +45,168 @@
                     android:visibility="gone"
                     tools:visibility="visible"/>
 
-                <LinearLayout
+                <net.cachapa.expandablelayout.ExpandableLayout
                     android:id="@+id/speakerInfoContainer"
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
-                    android:orientation="vertical"
-                    android:visibility="gone"
-                    tools:visibility="visible">
-                    <ImageView
-                        android:id="@+id/speakerAvatar"
-                        android:layout_gravity="center"
-                        android:layout_width="@dimen/avatar_large"
-                        android:layout_height="@dimen/avatar_large"
-                        android:padding="@dimen/padding_large"
-                        app:srcCompat="@drawable/ic_account_circle_grey" />
+                    app:el_duration="300"
+                    app:el_expanded="true"
+                    app:el_parallax="0.5">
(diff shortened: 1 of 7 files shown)


anhanh11001:
what kind of library should I look for?

iamareebjamal:
https://github.com/bravoborja/ReadMoreTextView

anhanh11001:
I didn't any use existing "ReadMoreTextView" library because it is unrelated to the issue and I also find it unnecessary to use a library for that

app/src/main/java/org/fossasia/openevent/general/utils/extensions/ViewExt.kt
```diff
@@ -24,3 +28,110 @@ fun View.showWithFading(duration: Long = 200L) {
...
+        }
+
+        override fun willChangeBounds(): Boolean {
+            return true
+        }
+    }
+
+    animation.duration = 300
+    startAnimation(animation)
+}
```
iamareebjamal: I find it unnecessary to write 100 lines of code to do something which is just an enhancement and unrelated to the core working of the project. It can impact performance, and contain bugs, we have to maintain it. 2 months later, nobody will understand the code and will not be able to maintain it. Same thing happened in server where instead of using a well-maintained library, they wrote celery integration by hand. New version of celery came and it broke. If they were using a library, it would have been a matter of bumping the version. That is why we use libraries

iamareebjamal:
> I also find it unnecessary to use a library for that

Trust me, if I had a library which did 2+2, I'd use that instead of writing 2+2. You don't know how hard it is to maintain code when it is not your core logic and a person who wrote it has stopped maintaining the project and code rot eventually happens and the code has bugs.

anhanh11001:
updated

app/src/main/res/layout/fragment_speakers_call_proposal.xml
```diff
@@ -45,13 +45,17 @@
...
-                    android:visibility="gone"
-                    tools:visibility="visible">
+                    app:el_duration="300"
+                    app:el_expanded="true"
+                    app:el_parallax="0.5">
+                    <LinearLayout
+                        android:orientation="vertical"
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content">
                     <ImageView
```
iamareebjamal: Wrong indentation

Issue #2201, opened 2019-07-28
Profile Image Not Editable After Taking A New One


anhanh11001:
**Describe the bug**
Profile Image Not Editable After Taking A New One

**Screenshots**
<img src="https://i.ibb.co/s5ySfx8/ezgif-5-d184eeae9e57.gif" alt="ezgif-5-d184eeae9e57" border="0">

**Would you like to work on the issue?**
Yes

PR #2203, opened 2019-07-28
fix: New profile image editable

anhanh11001:
Details:
- Make new profile image taken editable by saving userAvatar

Fixes: #2201

Screenshots for the change:
ezgif-2-269599000f16

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/EditProfileViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/EditProfileViewModel.kt
@@ -30,6 +30,7 @@ class EditProfileViewModel(
     private var updatedImageTemp = MutableLiveData<File>()
     var avatarUpdated = false
     var encodedImage: String? = null
+    var userAvatar: String? = null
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/EditProfileViewModel.kt
```diff
@@ -30,6 +30,9 @@ class EditProfileViewModel(
     private var updatedImageTemp = MutableLiveData<File>()
     var avatarUpdated = false
     var encodedImage: String? = null
+    lateinit var userAvatar: String
+
+    fun userAvatarInitialized() = this::userAvatar.isInitialized
```
liveHarshit: Why?

anhanh11001: to retain userAvatar when the fragment is destroyed

liveHarshit: I think the null check can work here?

Issue #2255, opened 2019-08-08
App crashes on searching events in my location


anhanh11001:
**Describe the bug**
App crashes on searching events in my location when the Map API doesn't return any location near me

**Logs**
```
 java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at java.util.ArrayList.get(ArrayList.java:437)
        at org.fossasia.openevent.general.search.location.LocationServiceImpl$getAdministrativeArea$1$1.onLocationChanged(LocationServiceImpl.kt:33)
        at android.location.LocationManager$ListenerTransport._handleMessage(LocationManager.java:356)
        at android.location.LocationManager$ListenerTransport.-wrap0(Unknown Source:0)
        at android.location.LocationManager$ListenerTransport$1.handleMessage(LocationManager.java:236)
        at android.os.Handler.dispatchMessage(Handler.java:108)
        at android.os.Looper.loop(Looper.java:166)
        at android.app.ActivityThread.main(ActivityThread.java:7529)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:245)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:921)
```

**Screenshots**
<img src="https://i.ibb.co/NCDSpFt/ezgif-5-5242746fd0cc.gif" alt="ezgif-5-5242746fd0cc" border="0">

**Would you like to work on the issue?**
Yes

PR #2259, opened 2019-08-08
fix: App crash when no location found

anhanh11001:
Fixes: #2255

Screenshots for the change:
ezgif-5-64aa352a246e


liveHarshit:
Fix build

anhanh11001:
> Fix build

fixed

anhanh11001:
LocationServiceImpl file changed here is for fdroid build. I don't know why fdroid and release build has a different implementation on LocationServiceImpl. In release build, a listener to LocationCallback is used for choosing the location near me instead of LocationListener like fdroid @liveHarshit

liveHarshit:
@anhanh11001 We are using google services with location manager for play store build for fetching faster and for fdroid build, only location manager is used.

anhanh11001:
@liveHarshit fixed

Issue #2258, opened 2019-08-08
Button in AttendeeFragment show "Register" for paid orders


anhanh11001:
**Describe the bug**
Button in AttendeeFragment show "Register" for paid orders, it should show "Order now" button

**Screenshots**
<img src="https://i.ibb.co/v3C0BKp/67760362-955263534819338-3903026589364912128-n.jpg" alt="67760362-955263534819338-3903026589364912128-n" border="0">

**Would you like to work on the issue?**
Yes

PR #2260, opened 2019-08-08
fix: Add Pay Now button based on payment method

anhanh11001:
Details:
- Add Pay Now button based on the payment method

Fixes: #2258

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
@@ -278,12 +278,7 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
             setupPaymentOptions(currentEvent)
             loadEventDetailsUI(currentEvent)
         }
-
-        rootView.register.text =
-            if (rootView.stripeRadioButton.isSelected || rootView.paypalRadioButton.isSelected)
-                getString(R.string.pay_now)
-            else
-                getString(R.string.register)
+        rootView.register.text = if (safeArgs.amount > 0) getString(R.string.pay_now) else getString(R.string.register)
     }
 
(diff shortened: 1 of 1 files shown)


nikit19:
Looks good, just fix build

anhanh11001:
> Looks good, just fix build

Fixed @nikit19

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeFragment.kt
```diff
@@ -278,12 +278,7 @@ class AttendeeFragment : Fragment(), ComplexBackPressFragment {
             setupPaymentOptions(currentEvent)
             loadEventDetailsUI(currentEvent)
         }
-
-        rootView.register.text =
-            if (rootView.stripeRadioButton.isSelected || rootView.paypalRadioButton.isSelected)
-                getString(R.string.pay_now)
-            else
-                getString(R.string.register)
+        rootView.register.text = if (safeArgs.amount > 0) getString(R.string.pay_now) else getString(R.string.register)
```
liveHarshit: Why? How can you pay now if the user has selected an offline payment option??

liveHarshit: PAY NOW should display when the user is paying through an online payment mode. Otherwise, the user is only registering for the event.

anhanh11001: At first they haven't selected any payment option so I just kept it that way. When they select a payment method, register will be displayed instead of pay now

liveHarshit: Oh, sorry, get it :+1:

PR #2264, opened 2019-08-09
feat: Implement JWT refresh token for authorization

### LINKED ISSUE #2230: Implement refresh tokens authentication
https://root-m29xfdiov.now.sh/

liveHarshit:
Fixes #2230

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/StartupViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/StartupViewModel.kt
@@ -20,6 +20,10 @@ import org.fossasia.openevent.general.utils.extensions.withDefaultSchedulers
 import retrofit2.HttpException
 import timber.log.Timber
 
+const val HEADER_TYPE = "headerType"
+const val JWT_ACCESS_TOKEN = "jwtAccessToken"
+const val JWT_REFRESH_TOKEN = "jwtRefreshToken"
+
 class StartupViewModel(
     private val preference: Preference,
     private val resource: Resource,
@@ -114,6 +118,25 @@ class StartupViewModel(
                 Timber.d("Settings fetched successfully")
             }, {
                 Timber.e(it, "Error in fetching settings form API")
+            })
+    }
+
+    fun checkAccessToken() {
+        if (!authHolder.isLoggedIn() || authHolder.isTokenValid())
+            return
+        preference.putString(HEADER_TYPE, JWT_REFRESH_TOKEN)
+        compositeDisposable += authService.refreshToken()
+            .withDefaultSchedulers()
+            .doFinally {
+                preference.putString(HEADER_TYPE, JWT_ACCESS_TOKEN)
+            }.subscribe({
+                authHolder.accessToken = it.accessToken
+            }, {
+                if (it is HttpException) {
+                    if (it.code() == HttpErrors.UNAUTHORIZED)
+                        logoutAndRefresh()
+                }
+                Timber.e(it, "Error refreshing token, logged out.")
             })
     }
 }
--- a/app/src/main/java/org/fossasia/openevent/general/auth/AuthHolder.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/AuthHolder.kt
@@ -3,30 +3,52 @@ package org.fossasia.openevent.general.auth
 import org.fossasia.openevent.general.data.Preference
 import org.fossasia.openevent.general.utils.JWTUtils
 
-private const val TOKEN_KEY = "TOKEN"
+private const val ACCESS_TOKEN = "accessToken"
+private const val REFRESH_TOKEN = "refreshToken"
 
 class AuthHolder(private val preference: Preference) {
 
-    var token: String? = null
+    var accessToken: String? = null
         get() {
-            return preference.getString(TOKEN_KEY)
+            return preference.getString(ACCESS_TOKEN)
         }
         set(value) {
             if (value != null && JWTUtils.isExpired(value))
-                throw IllegalStateException("Cannot set expired token")
+                throw IllegalStateException("Cannot set expired accessToken")
             field = value
-            preference.putString(TOKEN_KEY, value)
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/fossasia/openevent/general/StartupViewModel.kt
```diff
@@ -116,4 +120,23 @@ class StartupViewModel(
...
+    fun checkAccessToken() {
+        if (!authHolder.isLoggedIn() || authHolder.isTokenValid())
+            return
+        preference.putString(HEADER_TYPE, JWT_REFRESH_TOKEN)
+        compositeDisposable += authService.refreshToken()
+            .withDefaultSchedulers()
+            .doFinally {
+                preference.putString(HEADER_TYPE, JWT_ACCESS_TOKEN)
+            }.subscribe({
+                authHolder.accessToken = it.accessToken
```
anhanh11001: so header type is still set to JWT_ACCESS_TOKEN even though `subscribe()` throws an error?

liveHarshit: Yes, it will logout the user due to the unauthorization error.

liveHarshit: Add we are using access token for all API calls

app/src/main/java/org/fossasia/openevent/general/StartupViewModel.kt
```diff
@@ -116,4 +120,23 @@ class StartupViewModel(
...
+            }, {
+                if (it is HttpException) {
+                    if (it.code() == HttpErrors.UNAUTHORIZED)
```
nikit19: These two if statements can be combined

liveHarshit: Can't, according to Kotlin smart cast.

iamareebjamal:
Will test and see

app/src/main/java/org/fossasia/openevent/general/auth/AuthService.kt
```diff
@@ -66,7 +65,7 @@ class AuthService(
 
     fun logout(): Completable {
         return Completable.fromAction {
-            authHolder.token = null
+            authHolder.accessToken = null
```
iamareebjamal: Remove refresh token as well

app/src/main/java/org/fossasia/openevent/general/auth/AuthHolder.kt
```diff
@@ -3,37 +3,59 @@ package org.fossasia.openevent.general.auth
...
         }
 
-    fun getAuthorization(): String? {
+    var refreshToken: String? = null
+        get() {
+            return preference.getString(REFRESH_TOKEN)
+        }
+    set(value) {
+        if (value != null && JWTUtils.isExpired(value))
+            throw IllegalStateException("Cannot set expired refreshToken")
```
iamareebjamal: I mean this will crash the app or what? User should be logged out in this case

app/src/main/java/org/fossasia/openevent/general/auth/RequestAuthenticator.kt
```diff
@@ -2,11 +2,23 @@ package org.fossasia.openevent.general.auth
...
+) : Interceptor {
 
     override fun intercept(chain: Interceptor.Chain): Response {
-        val authorization = authHolder.getAuthorization()
+        val headerType = preference.getString(HEADER_TYPE, JWT_ACCESS_TOKEN)
+        val authorization =
+            if (headerType == JWT_REFRESH_TOKEN)
+                authHolder.getRefreshAuthorization()
+            else
+                authHolder.getAccessAuthorization()
```
iamareebjamal: No, no, no. How does it refresh the token when it is expired?

See the ansers here https://stackoverflow.com/questions/22450036/refreshing-oauth-token-using-retrofit-without-modifying-all-calls

liveHarshit: Good approach. :+1:

liveHarshit:
While injecting auth service in the authorization, the app is crashing with module definition. Any idea why we can inject auth holder but not auth service?

liveHarshit:
Problem is how to do API call from the authenticator to refresh token, @iamareebjamal can I use view model in authenticator?

iamareebjamal:
See the comment in the same answer. There is a circular dependency

liveHarshit:
> See the comment in the same answer. There is a circular dependency

Are you talking about this: https://stackoverflow.com/a/51360214
Currently, I'm getting infinite calls.

iamareebjamal:
I meant don't use dependency injection there. 

Manually create a separate service

liveHarshit:
> Manually create a separate service

I have a doubt, it needs to create a different HTTP client or using the same dependency injection module.

iamareebjamal:
Different HTTP client

iamareebjamal:
Status?

aggarwalpulkit596:
If anyone isn't working i can I've already implemented the same in my project 😅

iamareebjamal:
Go ahead

aggarwalpulkit596:
@iamareebjamal i guess the flow will be like if i get 401 in response then i need to use the refresh token to get another jwt token and replace that token from the existing right?

iamareebjamal:
https://stackoverflow.com/questions/22450036/refreshing-oauth-token-using-retrofit-without-modifying-all-calls

aggarwalpulkit596:
thanks

aggarwalpulkit596:
the current mechanism uses an interceptor for attaching the jwt token should I make pr for that first to move it to authenticator ?

iamareebjamal:
Do it in one PR

aggarwalpulkit596:
@iamareebjamal yeah I got figured it out that we have to use interceptor by default and authenticator for refreshing the token

Issue #2256, opened 2019-08-08
Map in EventDetailsFragment showing nothing


anhanh11001:
**Describe the bug**
In event details fragment, sometimes map show nothing

**Screenshots**
<img src="https://i.ibb.co/sbQmHYP/67566482-2207007072761486-3548067273430269952-n.jpg" alt="67566482-2207007072761486-3548067273430269952-n" border="0">

**Would you like to work on the issue?**
Yes

liveHarshit:
Debug or production?

anhanh11001:
production, haven't tested in debug

liveHarshit:
I think it is because the organizer has not provided a proper location. you can check other events. It's working.

PR #2266, opened 2019-08-09
fix: Hide map when no information is present

anhanh11001:
Detail:
- When event doesn't contain any location details (latitude, longitude), imageMap is hide

Fixes: #2256


app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -459,8 +459,7 @@ class EventDetailsFragment : Fragment() {
         // load location to map
         val mapClickListener = View.OnClickListener { startMap(event) }
 
-        val locationNameIsEmpty = event.locationName.isNullOrEmpty()
-        if (!locationNameIsEmpty) {
+        if (!event.locationName.isNullOrEmpty() && event.longitude != null && event.latitude != null) {
```
liveHarshit: What if the only location name is there, the map should display, right?

Issue #2133, opened 2019-07-16
Stripe not working


liveHarshit:
Charge class is not matching with the server documentation.

For PayPal same endpoint can be used - 
![image](https://user-images.githubusercontent.com/27783786/61267378-5ba57500-a7b5-11e9-8920-a5d0aab9fae5.png)

**Would you like to work on the issue?**

yes

liveHarshit:
Stripe will not work like that. Here we are just creating a token from the server using card data, not completing the payment.
https://stackoverflow.com/a/50233021

liveHarshit:
For now, if we redirect to the frontend, stripe, and PayPal both will work with the same implementation. Later Braintree implementation can handle both.

iamareebjamal:
No, we need stripe to be implemented separately as an organizer may want to enable stripe on server and not braintree

anhanh11001:
>Stripe will not work like that. Here we are just creating a token from the server using card data, not completing the payment.

We do create the token in the client-side and then send the success token to the server.
<img src="https://i.ibb.co/ww8vrKy/stripe.png" alt="stripe" border="0">

Can you tell me which part of the code is not correct?

liveHarshit:
> Stripe will not work like that. Here we are just creating a token from the server using card data, not completing the payment.
> https://stackoverflow.com/a/50233021

@anhanh11001

liveHarshit:
It's only creating a token from the card data, not completing the payment.

anhanh11001:
Ok so here is what we want right:
Collect card data (client) -> Create a stripe token using Stripe SDK (client) -> send token to the server (client) -> make payment (server) -> Finallize payment
Is it correct?

Because it is what is implemented. When we send the Charge object to the server, the server will handle the payment, send us the response and then we finalize it. We don't use Stripe SDK to make the charge, but our own API to make the charge

liveHarshit:
> > make payment (server) ->

Yes but here, how the server is handling the payment, Where the user is verifying the card with OTP or protected password!

liveHarshit:
With the test key, the server is throwing an error and the same in the production app also.

iamareebjamal:
OK, open an issue about test key not working as expected and mention what you sent and what you expect in response

liveHarshit:
[...]

PR #2269, opened 2019-08-13
fix: Add error utils, fix stripe error message and charge data class

liveHarshit:
Part of #2133

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
@@ -24,6 +24,7 @@ import org.fossasia.openevent.general.order.OrderService
 import org.fossasia.openevent.general.settings.SettingsService
 import org.fossasia.openevent.general.ticket.Ticket
 import org.fossasia.openevent.general.ticket.TicketService
+import org.fossasia.openevent.general.utils.ErrorUtils
 import org.fossasia.openevent.general.utils.HttpErrors
 import retrofit2.HttpException
 import timber.log.Timber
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -354,7 +357,14 @@ class AttendeeViewModel(
                     Timber.d("Failed charging the user")
                 }
             }, {
-                mutableMessage.value = resource.getString(R.string.payment_not_complete_message)
+                if (it is HttpException) {
+                    val body = it.response()?.errorBody()
+                    val jsonObject = JSONObject(body?.string()).getJSONArray(ERRORS)[0]
```
iamareebjamal: This will crash the app if not present

liveHarshit: So I think, try-catch can be used.

iamareebjamal: Copy ErrorUtils from orga app and use getErrorDetails, so this can be handled once for everywhere

liveHarshit: Updated

app/src/main/java/org/fossasia/openevent/general/utils/ErrorUtils.kt
```diff
@@ -0,0 +1,122 @@
...
+        var error = Error()
+        if (throwable is HttpException) {
+            val errorCode = throwable.code()
+
+            error = if (errorCode == BAD_REQUEST ||
+                errorCode == UNAUTHORIZED ||
+                errorCode == FORBIDDEN ||
+                errorCode == NOT_FOUND ||
+                errorCode == METHOD_NOT_ALLOWED ||
+                errorCode == REQUEST_TIMEOUT) {
```
anhanh11001: Is there any problem using Kotlin when expression for this here, I think it will look cleaner

liveHarshit: Sorry, can't get it?

iamareebjamal: https://kotlinlang.org/docs/reference/control-flow.html#when-expression

liveHarshit: updated

PR #2287, opened 2019-08-18
fix: Handle unverified account error

### LINKED ISSUE #2286: Handle error for unverified email
**Describe the bug**
Redirect to profile if an unverified user tries to buy a free ticket. 

yes

liveHarshit:
Fixes #2286 
<p align="center">
 
</p>

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
@@ -42,6 +42,7 @@ const val PAYMENT_MODE_PAYPAL = "paypal"
 const val PAYMENT_MODE_STRIPE = "stripe"
 private const val ERRORS = "errors"
 private const val DETAIL = "detail"
+private const val UNVERIFIED_USER = "unverified user"
 private const val ORDER_EXPIRY_TIME = 15
 
 class AttendeeViewModel(
@@ -79,6 +80,8 @@ class AttendeeViewModel(
     val stripeOrderMade: LiveData<Boolean> = mutableStripeOrderMade
     private val mutableOrderExpiryTime = MutableLiveData<Int>()
     val orderExpiryTime: LiveData<Int> = mutableOrderExpiryTime
+    private val mutableRedirectToProfile = SingleLiveEvent<Boolean>()
+    val redirectToProfile = mutableRedirectToProfile
 
     val attendees = ArrayList<Attendee>()
     private val attendeesForOrder = ArrayList<Attendee>()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -165,6 +167,11 @@ class AttendeeViewModel(
                 mutablePendingOrder.value = it
                 orderIdentifier = it.identifier.toString()
             }, {
+                if (it is HttpException) {
+                    if (it.code() == HttpErrors.UNPROCESSABLE_ENTITY) {
+                        mutableRedirectToProfile.value = true
```
ShridharGoel: But this error can be due to other reasons too.

liveHarshit: Then what should be the solution?

ShridharGoel: Check the pointer field of the error and if it indicates unverified account error, then redirect the user to profile.

liveHarshit: {status: 422, source: {pointer: "/data/relationships/order"}, title: "Unprocessable Entity",…}

ShridharGoel: And it does not have any message?

liveHarshit: I'm not sure if it is okay to compare the message: detail: "Unverified user cannot place free orders"

ShridharGoel: @iamareebjamal I think there should have been more information in the pointer.

iamareebjamal: Check if message containsIgnoreCase 'unverified user' for now. Open an issue on server to change the error to `Unauthorized` from `Unprocessable Entity` and correct the pointer. Both are wrong

liveHarshit: Updated

app/src/main/java/org/fossasia/openevent/general/attendees/AttendeeViewModel.kt
```diff
@@ -165,6 +168,11 @@ class AttendeeViewModel(
                 mutablePendingOrder.value = it
                 orderIdentifier = it.identifier.toString()
             }, {
+                if (it is HttpException) {
+                    if (ErrorUtils.getErrorDetails(it).detail?.contains(UNVERIFIED_USER, true) == true) {
```
iamareebjamal: containsIgnoreCase?

liveHarshit: > containsIgnoreCase?

Ignore case is set to true. `contains(UNVERIFIED_USER, true)`

Issue #2299, opened 2019-08-23
Save attendees to db when fetching order


anhanh11001:
**Describe the feature you'd like**
When fetching order, attendees should also be fetched and saved to db

**Additional context**
https://github.com/fossasia/open-event-attendee-android/pull/2281

**Would you like to work on the issue?**
Yes

PR #2281, opened 2019-08-18
feat: Save tickets to DB

anhanh11001:
Fixes: #2205

**Screenshots**
ezgif-5-dc6d209eba11

**Related Context**
Work left: Fetch attendees and save to database when fetching order (https://github.com/fossasia/open-event-attendee-android/issues/2299)

PR #2300, opened 2019-08-23
feat: Save attendees to DB

anhanh11001:
Fixes: #2299


app/src/main/java/org/fossasia/openevent/general/order/OrderDetailsFragment.kt
```diff
@@ -144,8 +121,44 @@ class OrderDetailsFragment : Fragment() {
...
+                    .error(R.drawable.header)
+                    .placeholder(R.drawable.header)
+                    .into(rootView.backgroundImage)
+            })
+
+        orderDetailsViewModel.attendees
+            .nonNull()
+            .observe(viewLifecycleOwner, Observer {
+                if (it.isEmpty()) {
+                    Toast.makeText(context, getString(R.string.error_fetching_attendees), Toast.LENGTH_SHORT).show()
```
liveHarshit: Sanckbar

anhanh11001: > How you're revalidating attendees form API?

I don't think it is necessary to revalidate attendees from API as it is not possible to update attendees after order is made. If we need to then it is still possible as in https://github.com/fossasia/open-event-attendee-android/pull/2281 SwipeRefreshLayout is added to refresh order, which should also refresh attendees if both PRs got merged.

PR #2353, opened 2019-09-19
Migrate to Gradle Kotlin DSL

### LINKED ISSUE #1410: [Maintainence] Make Dependencies file for maintaining all dependencies/libraries
**Describe the chore**
Make a separate `dependencies` file to list all dependencies version number / libraries versions and reference those versions in `build.gradle` files

This would include `dependencies.gradle` file which would contain all versions of dependencies/ libraries used . This would help us in maintaining/ adding the future dependencies . 

- [ :heavy_check_mark: : ] Yes
- [ ] No
- Other:

Skrilltrax:
Fixes #1410 

Changes: Converted gradle files to kotlin dsl syntax


buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -0,0 +1,145 @@
...
+ * YOU are responsible for updating manually the dependency version.
+ */
+object Versions {
+    const val com_diffplug_gradle_spotless_gradle_plugin: String = "3.24.2"
+
+    const val de_fayard_buildsrcversions_gradle_plugin: String = "0.6.1"
+
+    const val navigation_safe_args_gradle_plugin: String = "1.0.0"
+
+    const val com_android_tools_build_gradle: String = "3.5.0"
```
liveHarshit: No need for "String"

Skrilltrax: I can remove it for now but since it is an auto generated file, every time you will check for dependency updates it'll just add it back.

liveHarshit: If it is auto-generated, let it be same.

Skrilltrax:
@liveHarshit do i need to resolve codacy issues because most of the UnsafeCast are due to the return type being Any for the functions.

liveHarshit:
> @liveHarshit do i need to resolve codacy issues because most of the UnsafeCast are due to the return type being Any for the functions.

You can remove it by using smart cast and removing non-null assertions.

iamareebjamal:
First goal should be passing build. Second would be polishing because the current library declarations are too verbose. We only want versions for libraries which have duplicate dependencies. We also don't know if dependabot supports this format. We can't lose auto dependency update functionality

adityastic:
@iamareebjamal dependabot currently does not support kts. The issue is opened since a year though https://github.com/dependabot/feedback/issues/172

Skrilltrax:
@iamareebjamal @adityastic so should I work on kts or close this PR and check groovy dependency management ?

iamareebjamal:
Leave it be. We'll iterate on it when the support is added

Skrilltrax:
@iamareebjamal sure, I'll just update the PR so to make sure it passes all the checks.

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -0,0 +1,145 @@
...
+ * This will only update the comments.
+ *
+ * YOU are responsible for updating manually the dependency version.
+ */
+object Versions {
+    const val com_diffplug_gradle_spotless_gradle_plugin: String = "3.24.2"
+
+    const val de_fayard_buildsrcversions_gradle_plugin: String = "0.6.1"
+
+    const val navigation_safe_args_gradle_plugin: String = "1.0.0"
```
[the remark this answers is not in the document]
Skrilltrax: It is an auto-generated file

app/build.gradle.kts
```diff
@@ -0,0 +1,256 @@
...
+    // Searchable Spinner
+    implementation(Libs.searchablespinnerlibrary)
+
+    //ExpandableTextView
+    implementation(Libs.expandabletextview)
+
+    //ExpandableView
+    implementation(Libs.expandablelayout)
+
+
```
nikit19: extra

Skrilltrax: fixed

PR #2391, opened 2019-10-13
fix: Similar events collapsing

### LINKED ISSUE #2390: Similar events layout collapses at end
**Describe the bug**
In the EventDetailFragment, the similar events layout collapses at the end of the pagination. 

**To Reproduce**
1. Go to the Events tab.
2. Click on any event in the recyclerview.
3. Scroll down to the end for the "More like this" section.
4. Swipe right till the end of the pagination.

**Expected behavior**
The layout should be gone only when there are no similar events on the list. In the rest of the cases, it should be there in the UI.

**Screenshots**
ezgif-1-e627d8f5a55c

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | Redmi Note 5 |
|Android Version|  Oreo 8.1 |

- [x] Yes
- [ ] No
- Other:

tronku:
Fixes #2390 

Changes: 
Previously the visibility of the container layout was dependent on the size of the list fetched from the API. So, whenever the list reaches the end of the pagination, due to size 0, the visibility of the layout was set as GONE. Now, I have used a boolean hasSimiliarEvents (initially false) which get updated depending on the size of the list and container layout's visibility depends on this variable. So, even if there are events on any page the variable will be set as true otherwise false. This resolved the issue.

Screenshots for the change:
ezgif-1-c507beb61fa8

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -387,7 +387,7 @@ class EventDetailsFragment : Fragment() {
                     if (!similarEventsAdapter.currentList.isNullOrEmpty()) {
                         hasSimilarEvents = true
                     }
-                    rootView.similarEventsContainer.isVisible =  hasSimilarEvents
+                    rootView.similarEventsContainer.isVisible = hasSimilarEvents
                 }
             })
 
--- a/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
@@ -103,6 +103,7 @@ class EventDetailsFragment : Fragment() {
     private val sessionsAdapter = SessionRecyclerAdapter()
     private val socialLinkAdapter = SocialLinksRecyclerAdapter()
     private val similarEventsAdapter = SimilarEventsListAdapter()
+    private var hasSimilarEvents: Boolean = false
 
     private lateinit var rootView: View
     private lateinit var binding: FragmentEventBinding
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -383,7 +384,10 @@ class EventDetailsFragment : Fragment() {
                     rootView.similarEventsContainer.isVisible = true
                 } else {
                     rootView.shimmerSimilarEvents.stopShimmer()
-                    rootView.similarEventsContainer.isVisible = similarEventsAdapter.currentList?.isEmpty() ?: true
+                    if (!similarEventsAdapter.currentList?.isEmpty()!!) {
```
liveHarshit: do not use non-null assertions

tronku: Resolved

app/src/main/java/org/fossasia/openevent/general/event/EventDetailsFragment.kt
```diff
@@ -383,7 +384,10 @@ class EventDetailsFragment : Fragment() {
                     rootView.similarEventsContainer.isVisible = true
                 } else {
                     rootView.shimmerSimilarEvents.stopShimmer()
-                    rootView.similarEventsContainer.isVisible = similarEventsAdapter.currentList?.isEmpty() ?: true
+                    if (!similarEventsAdapter.currentList.isNullOrEmpty()) {
+                        hasSimilarEvents = true
+                    }
+                    rootView.similarEventsContainer.isVisible =  hasSimilarEvents
```
nikit19: fix indentation

tronku: what's the issue with indentation?

PR #2469, opened 2019-12-09
fix: Issue #2468 by removing first element of arraylist

### LINKED ISSUE #2468: Empty String in Popular Location
**Describe the bug**
The first option of popular location is an empty string. Due to which on selecting that option, all the events of all the locations are shown

**To Reproduce**
1. Go to 'Enter location' in events screen
2. Under popular locations, the first option is an empty string. 

**Expected behavior**
The empty string should be removed from that list. Because that string is causing the app to behave unexpectedly.

**Logs**

**Screenshots**

This screenshot shows the empty string in the list.
WhatsApp Image 2019-12-09 at 12 05 58 AM

This screenshot shows the unexpected behaviour of the app as all the events are coming.
WhatsApp Image 2019-12-09 at 12 09 23 AM

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Redmi Note 5|
|Android Version       |Oreo 8.1|

**Additional context**

- [✓ ] Yes
- [ ] No

VarunSaini23:
Resolve #2468 

Changes: Removed the first element from the ArrayList as it was representing an empty string. So everything is fine now. The app behaves as expected.

Screenshots for the change:

WhatsApp Image 2019-12-09 at 11 08 09 PM

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/search/location/LocationAdapter.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/search/location/LocationAdapter.kt
@@ -12,6 +12,7 @@ class LocationsAdapter : RecyclerView.Adapter<LocationViewHolder>() {
     fun addAll(locations: List<String>) {
         if (locationsList.isNotEmpty()) locationsList.clear()
         locationsList.addAll(locations)
+        locationsList.removeAt(0);
         notifyDataSetChanged()
     }
 


VarunSaini23:
> It looks like a workaround.
Basically the first item from the backend api is an empty string. Need to change it in the backend.
API endpoint - https://open-event-api-dev.herokuapp.com/v1/event-locations?sort=name

But till that time, this workaround will work good

iamareebjamal:
No, it won't work good. It's a bug in backend

VarunSaini23:
> No, it won't work good. It's a bug in backend, not in the app

So it should be reported to the backend team?

iamareebjamal:
Yes

VarunSaini23:
> Yes

Okay. I will do that

PR #2484, opened 2019-12-10
fix: branch release for open-event-attendee

### LINKED ISSUE #2483: Fix Branch Release

adityastic:
Fixes #2483

### DIFF
--- a/scripts/update-apk.sh
+++ b/scripts/update-apk.sh
@@ -60,10 +60,9 @@ for file in app*; do
 done
 
 if $IS_PUBLISH_BRANCH ;then
-    cd ..
     gem install fastlane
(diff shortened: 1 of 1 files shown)


adityastic:
@iamareebjamal should fix the branch release

scripts/update-apk.sh
```diff
@@ -69,6 +69,7 @@ if $IS_PUBLISH_BRANCH ;then
     if $PR_FOR_RELEASE ;then
         exit 0
     fi
+    cd apk
```
iamareebjamal: Can't we just remove `cd ..` above and change paths accordingly

adityastic: @iamareebjamal i didn't get what you meant by changing paths accordingly

iamareebjamal: The path of apk and fastlane json

adityastic: Oh yes, we can definitely do that. Let me update the pr

PR #2492, opened 2019-12-10
fix: Resolve password reset mail issue

### LINKED ISSUE #2476: UI- Forgot-password_notification
1. When we press "I forgot my password" in login there is no pop-up notification regarding the password reset link sent to the respective email id.

To Reproduce the first bug;
1. Go to 'Profile'
2. Click on 'login'
3. Enter the registered email and continue
4. See error -->When you press "I forgot my password, it shows loading and no notification 
    appears regarding the reset link sent.

**Expected behavior**
We expect a notification saying that "A reset link has been sent to your respective e-mail"

**Screenshots**
Screenshot_bug1

**Smartphone Info:**
|Device         |Redmi Note-5 Pro|
|Android Version|Android-10|

- [No]

tronku:
Fixes #2476 

Changes: Due to the difference in message being checked, mail sent layout was not visible. There were extra spaces in the response message so I have used regex to remove all spaces and checked accordingly.

Screenshot: 
WhatsApp Image 2019-12-11 at 12 41 46 AM

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/general/auth/LoginViewModel.kt
+++ b/app/src/main/java/org/fossasia/openevent/general/auth/LoginViewModel.kt
@@ -91,12 +91,12 @@ class LoginViewModel(
                 mutableRequestTokenSuccess.value = verifyMessage(it.message)
             }, {
                 mutableRequestTokenSuccess.value = verifyMessage(it.message.toString())
-                mutableError.value = resource.getString(R.string.email_not_in_server_message)
+                mutableError.value = resource.getString(R.string.reset_password_mail_message)
             })
     }
 
     private fun verifyMessage(message: String): Boolean {
-        if (message == resource.getString(R.string.email_sent)) {
+        if (message.replace(" ", "") == resource.getString(R.string.reset_password_mail_message)?.replace(" ", "") ?: "") {
             return true
         }
         return false
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,8 +23,8 @@ android {
         applicationId "com.eventyay.attendee"
         minSdkVersion 21
         targetSdkVersion 28
-        versionCode 16
-        versionName "0.9.0"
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/fossasia/openevent/general/auth/LoginViewModel.kt
```diff
@@ -91,12 +91,12 @@ class LoginViewModel(
...
             }, {
                 mutableRequestTokenSuccess.value = verifyMessage(it.message.toString())
-                mutableError.value = resource.getString(R.string.email_not_in_server_message)
+                mutableError.value = resource.getString(R.string.reset_password_mail_message)
             })
     }
 
     private fun verifyMessage(message: String): Boolean {
-        if (message == resource.getString(R.string.email_sent)) {
+        if (message.replace(" ", "") == resource.getString(R.string.reset_password_mail_message)?.replace(" ", "") ?: "") {
```
iamareebjamal: Show the message from server. This method is bound to fail whenever the message from server is changed

tronku: That's why I was asking you but you said that it's the app's problem.
Response - 
response

iamareebjamal: It is

iamareebjamal: I'm saying that show the {"message"} from the server in the app. Don't match the message from the server. Just unconditionally show it

tronku: I think there's no need of verifyMessage() function then.

iamareebjamal: Yes

PR #2493, opened 2019-12-10
chore: cd to prev dir for fastlane metadata

iamareebjamal:
During the release process, metadata will not get 
uploaded if we are in apk directory


adityastic:
There is a better way without using cd , should I create a pr?

iamareebjamal:
OK

PR #2571, opened 2020-01-18
chore: Add Chinese (Simplified) Language!

pranavkarthik10:
Fixes #2570 

Changes: 
Added the Chinese simplified language for many users whose primary language is this. This was done using Google Translate and with lots of checking from my friends and a teacher who's native language is Chinese.

Screenshots for the change:
Screenshot_20200117-225951_Eventyay Attendee
Screenshot_20200117-230019_Eventyay Attendee
Screenshot_20200117-230012_Eventyay Attendee

PR #2576, opened 2020-01-20
Revert "Add Chinese (Simplified) Language! (#2571)"

scixor:
Reverts fossasia/open-event-attendee-android#2571

As said by @cweitat in pull request https://github.com/fossasia/open-event-attendee-android/pull/2571 , this doesn't have perfect meaning of translation.. and should thereby be reverted 😇


iamareebjamal:
I feel the same way for all the translations added. There is no point of adding translations by copy pasting from Google Translate. I would not merge any translations, but unfortunately, it is a GCI task. It is better for the app to be untranslated than to have wrong translations

PR #2568, opened 2020-01-16
chore: Add support for thai language

aryanguls:
Fixes #2514 

Changes: For the purpose of this pull request I have added Thai language as a language of use to the Open Event Attendee App so that viewers may also view the content of the app in thai language and are not limited by the lack of options of languages available on the app.

@mariobehling @hpdang @nikit19 @liveHarshit 
Mentors please review and merge as and when possible. Thank you.

PR #2577, opened 2020-01-20
Revert "chore: Add support for thai language (#2568)"

hoangdatngo:
Revert #2568 
As said by @cweitat , this doesn't have perfect meaning of translation.. and should be reverted


hoangdatngo:
@iamareebjamal, @cweitat, @nikit19 Can any mentor please review it?

iamareebjamal:
Where was it said about thai translations. If you are thai, and you find translations to be wrong, then I'll merge it

PR #2579, opened 2020-01-20
revert: "chore: Add support for thai language"

scixor:
Reverts fossasia/open-event-attendee-android#2568

As per recent event while merging pr's  in https://github.com/fossasia/open-event-attendee-android/pull/2571...As mentioned by @cweitat there is no verification of this being correct so... should be reverted, as there no grantee of correct language translation 

Thanks


iamareebjamal:
Already reverted

PR #2596, opened 2020-01-22
fix: Changed edit profile's link based social system

### LINKED ISSUE #2526: Instagram link in profile.
There is some issue with the addition of Instagram link in profile. Once the data is entered and the profile is updated and the profile is opened to be edited again, the Instagram link can't be seen like the rest of the name and phone number once the edit pro

file option is clicked on.

### LINKED ISSUE #2590: Instagram Profile link update not working
**Describe the bug**
Whenever the user inputs a instagram profile link and updates it.. the updates doen't get changed

**To Reproduce**

1. Go to profile
2. Click on edit profile
3. Enter instagram link and click on update
4. Come back to the edit profile to see Instagram link isn't updated 

**Expected behavior**
Instagram link should update in user profile

**Screenshots**

**Smartphone Info:**
Any mobile 

- [ ] Yes
- [x] No

### LINKED ISSUE #2591: Entering username instead of links instead of links in profile edit
**Describe the issue**
This isn't really a bug but is a user clarification issue, currently in edit profile page, the user needs to enter the entire link with the protocol itself like `https` etc. which is tiresome to add in mobile as people using instagram in mobile probably use it in instagram app, not to mention chromium engines are now not showing the protocols in the address bar in general. This led to some of the issues like.

https://github.com/fossasia/open-event-attendee-android/issues/2524
https://github.com/fossasia/open-event-attendee-android/issues/2526
https://github.com/fossasia/open-event-attendee-android/issues/2525

**Expected behavior**
User Should be able to just type the username instead of the link

- [ ] Yes
- [x] No
- Other:

scixor:
- Fixes #2591 
- Fixes #2590 

## Changes:
- changed the way users add the links to use usernames instead of adding an entire link
- now making changes to Instagram link works (it was a one line change)
- this doesn't change the way we interface from the server but we just hides it from the User

## How this is done
- when we retrieve link from the server to display we strip off the protocol part by using this regEx pattern `((https)|(http)|[/|:]|(.com)|(www.)|(instagram)|(facebook)|(twitter))` which just leaves the username part which is then shown to the screen
- when user clicks on update button the input is taken from the fields and we run it through another regEx pattern `[$\s@,]` to remove any sort of extra things like space @ , ... and then add it to the back of a header URL.

you can check that expression here by pasting into https://www.regextester.com/ the regular expression box and changing the input text

Video for the change:
https://photos.app.goo.gl/FWfMLu377WTStAbSA

APK:
[app-fdroid-debug.zip](https://github.com/fossasia/open-event-attendee-android/files/4099766/app-fdroid-debug.zip)


app/src/main/res/values/strings.xml
```diff
@@ -432,7 +432,10 @@
     <string name="profile_settings">Profile Settings</string>
     <string name="loading_tickets">Loading tickets...</string>
     <string name="download_invoice">Download Invoice</string>
-    <string name="invalid_url_message">This is not a valid URL</string>
+    <string name="invalid_url_message">Please enter the correct username</string>
+    <string name="facebookHeadUrl">https://www.facebook.com/</string>
+    <string name="twitterHeadUrl">https://www.twitter.com/</string>
+    <string name="instagramHeadUrl">https://www.instagram.com/</string>
```
iamareebjamal: These are untranslatable strings and shouldn't be present here

iamareebjamal:
Add unit tests for extraction logic

scixor:
@iamareebjamal there seems to be a bigger issue... can you please check this with the play store version about this... when you add a social media link.. you can't remove it later.. I was thinking it was my fault from last 3 hours.. seems playstore version also has this

app/src/main/res/values/strings.xml
scixor: where should I keep them then?

scixor: I am adding `translatable="false"` tag to them and moving them at the bottom

scixor:
Also I have added a video in the description please check that too

scixor:
Added tests for regEx patterns

app/src/main/res/values/strings.xml
```diff
@@ -535,4 +535,8 @@
     <string name="tickets_for">Ticket(s) for %1$s</string>
     <string name="error_making_paypal_payment_message">Fail on making Paypal payment</string>
 
+    <!-- Non-Translatable -->
+    <string name="facebookHeadUrl" translatable="false">https://www.facebook.com/</string>
+    <string name="twitterHeadUrl" translatable="false">https://www.twitter.com/</string>
+    <string name="instagramHeadUrl" translatable="false">https://www.instagram.com/</string>
```
iamareebjamal: snake case

scixor: oh.. I thought we used Camel case.. I will fix this

iamareebjamal:
We want to show error and not silently convert usernames with spaces to without them

scixor:
i think we can add a warning saying... spaces will be removed automatically when updated or something? would that work?

scixor:
Done all the changes.. I have added a text view which pops up slowly mentioning this .. and also changed Update button to different color ... for Identifying that there might be an issue... I didn't add red because it didn't match with our cool bluish theme, and also this is just a warning.. 😀

scixor:
@iamareebjamal @liveHarshit linting is now also fixed 😀builds should pass now in Travis

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
```diff
@@ -32,17 +32,7 @@ import kotlinx.android.synthetic.main.dialog_edit_profile_image.view.editImage
...
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.firstName
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.instagram
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.lastName
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.phone
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.profilePhoto
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.profilePhotoFab
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.toolbar
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.twitter
-import kotlinx.android.synthetic.main.fragment_edit_profile.view.updateButton
+import kotlinx.android.synthetic.main.fragment_edit_profile.view.*
```
liveHarshit: Remove wildcard import.

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
```diff
@@ -0,0 +1,18 @@
...
+class RegExTest {
+    val frag = EditProfileFragment()
+    @Test
+    fun `protocol removal when loaded`() {
+        assertEquals("abcdefg", frag.loadStringRegEx("https://instagram.com/abcdefg"))
+    }
+
+    @Test
+    fun `anomaly removal when updating`() {
+        assertEquals("helloworld", frag.updateStringRegEx("h@ello@wo ,   rld"))
```
liveHarshit, line 16: it's invalid!

liveHarshit (changes requested):

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
scixor: @liveHarshit  why? I am new to the tests 😇 can you please explain?

app/src/main/java/org/fossasia/openevent/general/auth/EditProfileFragment.kt
scixor: I will fix that

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
scixor: it tests the function's use of removing the anomalies... I don't get the problem 😅

scixor:
@liveHarshit made the changes asked for.. other than test part.

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
scixor: I can't use that function to make entire links because .. it is also being used by change listener to analyze whether the text is right or not...

app/src/main/res/layout/fragment_edit_profile.xml
```diff
@@ -152,13 +154,26 @@
...
+            android:id="@+id/error"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:background="@color/black"
+            android:textColor="@color/white"
+            android:textSize="@dimen/text_size_medium"
+            android:padding="@dimen/padding_medium"
+            android:visibility="gone"
+            android:textAlignment="center"
+            android:text="@string/error_text" />
```
iamareebjamal: Remove this and use TextInput error

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
iamareebjamal, line 16: Since this is invalid username, it should throw. We are only going to make benign changes like URLs to username conversions and removing of `@` sign and it should be visible to user before they click save

app/src/main/res/layout/fragment_edit_profile.xml
scixor: sure

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
scixor: but invalid username is confirmed at a different function 😅 that function decides if the username is invalid or not.. this is just a regEx pattern function which takes a string removes any anomaly and return a string **(tests if regEx pattern removes any essential things)** .. I will make a new test case for invalid username

iamareebjamal: It should not do this. This is an invalid conversion

scixor: @iamareebjamal so how do we check and change the UI with onChangeListner and remove things when user Clicked on update with a same function 😕 ? this function is being used when updating and also while checking. I added that test if someone changes the pattern and if it leads to deleting of wrong characters .. now I also added another test...

Sid911:
Fixed most of the things.. 😀

app/src/test/java/org/fossasia/openevent/general/RegExTest.kt
scixor: I think this solves this conversation

scixor:
well if someone gets time.. please review this @iamareebjamal @liveHarshit @nikit19

scixor:
@liveHarshit please review 😃... thanks

scixor:
closing pr! because of no response whatsoever on this pr by maintainers 🤷‍♂️.. Maybe someone else could fix it.

PR #2787, opened 2020-10-10
fix: Fixed the UI glitch that happens when the events fragment is refreshed.

### LINKED ISSUE #2786: A UI glitch that happens in the events fragment when refreshed.
When the events fragment is refreshed ,  it can be clearly seen that , before the events are loaded i.e. after the shimmer, the no-events placeholder 
comes into place for a second before the events are displayed. Which hinders the user experience .

To reproduce the behavior 
Go to the events fragment  and swipe to refresh.

Expected behavior
The events should load immediately after the shimmer without displaying the no-event placeholder.

Smart Phone Info
Moto E5 plus
Pixel 2
Samsung M duo

I would like to work on this issue

VamsiKrishnaCommits:
Fixes #2786 

Changes: 
Now , on refresh , events are loaded right after the shimmer without the no-event placeholder coming in .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ cache:
       - "${TRAVIS_BUILD_DIR}/gradle/wrapper/dists/"
       - "$HOME/.gradle/caches/"
       - "$HOME/.gradle/wrapper/"
+
+licenses:
+  - '.+'
+
 before_script:
   - bash scripts/prep-key.sh
 script:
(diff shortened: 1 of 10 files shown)


liveHarshit (changes requested):
I can see the issue is closed (please add gif for before and after), and there are unnecessary changes, revert them

VamsiKrishnaCommits:
> I can see the issue is closed (please add gif for before and after), and there are unnecessary changes, revert them

Yes will do that

PR #2833, opened 2021-01-20
fix: incomprehensible snack bar message in the event details fragment

### LINKED ISSUE #2832: SnackBar showing incomprehensible pop out message in the EventDetails Fragment.
**Describe the bug**
The event I clicked was CodeHeat Ask me Anything January , somehow it failed to fetch something and popped out the below snack bar which does not make sense
**To Reproduce**
Click on the event CodeHeat AMA- January edition

**Expected behavior**
A sensible message 

**Screenshots**

Screenshot_20210119-211156_Eventyay Attendee

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Samsung  J7 duo|
|Android Version| Oreo  |

- [.] Yes

VamsiKrishnaCommits:
Fixes #2832 

Fixed issue that is resulting in the incorrect message. The issue was in the getstring method that merges two strings given appropriate place holders. So instead of a string , an array of strings were passed as the second argument on which the toString() function is invoked resulting in that alien looking expression.
Screenshots for the change:
Screenshot_20210120-234907_Eventyay Attendee

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -152,19 +152,19 @@ dependencies {
     def roomVersion = "2.2.5"
     def ktx_version = "1.1.0"
     def ktx2_version = "2.2.0"
-    def nav_version = "2.3.0"
+    def nav_version = "2.3.1"
     def anko_version = "0.10.8"
     def paging_version = "2.1.2"
 
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'androidx.multidex:multidex:2.0.1'
     implementation 'androidx.appcompat:appcompat:1.2.0'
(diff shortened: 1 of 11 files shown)


VamsiKrishnaCommits:
@iamareebjamal please review this

iamareebjamal:
Error shouldn't be shown

VamsiKrishnaCommits:
> Error shouldn't be shown

I think the social links are not present in the database for that particular event and is the reason it is returning nothing, that is what I found in the logs. What do you think? @iamareebjamal

iamareebjamal:
Yes

